PR #11, opened 2023-08-13
#8 support clang format & clang tidy

fktn-k:
This library now supports clang-format & clang-tidy.  
clang-format-16 & clang-tidy-16 are checked in my develop environment on Windows10 Home edition.  
Other versions may be currently unsupported, so further investigation may be necessary.  

clang-format is also supported while executing GitHub Actions, but clang-tidy is not.  
As for clang-tidy, implementation will be done in another issue later on.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,222 @@
+---
+Language:        Cpp
+# BasedOnStyle:  Microsoft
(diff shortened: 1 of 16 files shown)


fktn-k:
close #8

PR #131, opened 2023-10-09
#128 moved internal impl to detail dir/namespace

fktn-k:
To keep consistency of public fkYAML library APIs as much as possible, implementations for internal use only have been moved to `include/fkYAML/detail` directory.  
Such implementations are subject to changes without notices, and thus have no gurantee of consistency even if the fkYAML library provides its backward compatibility.  
Those moved implementations will still be listed in the fkYAML release package because they are still necessary for a project which uses the fkYAML library.  

Furthermore, part of doxygen comments have been fixed to make it up to day.  

See #128 for more details.


fktn-k:
To fix failure in GA workflow, the command for clang-format has been moved from CMakeLists.txt to Makefile.  
This would be reasonable since clang-format does not require build options for a header-only library.

PR #146, opened 2023-10-15
#133 refactor lexer

fktn-k:
In this PR, the `lexical_analyzer` class has been refactored by:
- implementing the `input_handler` class as a handler of an input adapter,
- implementing `from_string` template functions to generalize conversions from string tokens to native data type values,
- and deleting duplicate conversion checks for resulting token values.

Furthermore, some small fixes have also been applied such as typos in test case labels or splitting some too complicated functions into somewhat understandably sized ones.  

With the changes described above, backward compatibility is still guranteed since previous v0.1 releases.

### DIFF
--- a/include/fkYAML/detail/conversions/from_string.hpp
+++ b/include/fkYAML/detail/conversions/from_string.hpp
@@ -82,6 +82,25 @@ inline bool from_string(const std::string& s, type_tag<bool> /*unused*/)
 }
 
 /**
+ * @brief Specialization of from_string() for long values with std::string.
+ *
+ * @tparam  N/A
+ */
+template <>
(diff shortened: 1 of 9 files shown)


fktn-k:
TODOs for myself:  
- [ ] Add doxygen comments for input_handler class and from_string template functions
- [ ] Use std::strtoxxx() for (unsigned) int/long wgivh may increase performance

fktn-k:
> TODOs for myself:
> 
> * [x]  Add doxygen comments for input_handler class and from_string template functions
> * [x]  Use std::strtoxxx() for signed/unsigned long types may increase performance

I changed implementation to use std::stoxxx() instead of std::strtoxxx() to minimize differences among platforms/compilers in how to handle/express error cases.

PR #176, opened 2023-10-22
#175 support detecting indentation in deserialization

fktn-k:
Before this PR, the fkYAML library couldn't detect indentation width before each token of input buffers.  
Which resulted in wrong deserialization outputs with the following YAML formatted string:  
  ```yaml
  foo:
    bar: 123
  baz: true   # This was the part of "foo" mapping node.
  ```
This PR has corrected the above issue and support multiple mapping nodes at different levels.


fktn-k:
Though I have no idea why Codacy check is failing, I'll merge this PR since the other checks all passes successfully.

PR #178, opened 2023-10-22
#177 Add test for input adapters only

fktn-k:
Before this PR, there were no test cases only for input adapters.
To ensure that input adapters work as they are expected with no dependencies on the other classes, new test cases for input adapters have been added with some test data files.  

Along with the above implementation, some unused members in input adapter classes have been deleted.


fktn-k:
CTest failed in reading a test data file in the unit test cases for input adapters, just with MSVC (not with gcc/clang).  
So I made a quick workaround to specify the test data directory path by creating a header file in `test/unit_test/CMakeLists.txt` file.

Issue #292, opened 2024-03-21
Embedded flow indicators in plain unquoted scalar values cause incorrect parse results


stephenwhittle:
### Description

This issue is similar to #275 , but concerns values rather than keys. 
A flow sequence or mapping indicator is currently treated by the parser as the end of a plain unquoted scalar value, even outside of a flow collection.

### Reproduction steps

The following minimal YAML demonstrates the issue:
```yaml
KEY:
          description: Cryptographic key, bits [159 :128]
```

### Expected vs. actual results

The entire string `Cryptographic key, bits [159 :128]` should be treated as the value, but because the lexer treats [ as the end of a string, the resulting tree contains a key of 159 and a value of 128 with the rest of the string being discarded.

In my testing, checking `m_flow_context_depth` before treating those indicators as string terminators in the lexer, seems to address this issue.

### Compiler and operating system

VS 2022 (both clang-cl and msvc compilers), Win 10

### Library version

04e1a7b9115398ae8c57c0f5263dcb8390e3dc32

### Validation

- [X] The bug also occurs if the latest version from the [`develop`](https://github.com/fktn-k/fkYAML/tree/develop) branch is used.
- [X] I can successfully [compile and run the unit tests](https://github.com/fktn-k/fkYAML#how-to-execute-the-unit-tests).

fktn-k:
@stephenwhittle   
Thanks for the bug report!  
I overlooked that part while fixing the issue you referred in the description.  

Also, I totally agree with your solution and thanks again for the PR.  
I'll take a look soon.

PR #293, opened 2024-03-21
#292 Better handling for flow indicators in permitted scalar contexts

stephenwhittle:
This PR addresses #292 by checking if the lexer is in a flow context before treating flow indicators, commas, or colons as terminating plain unquoted strings.
---

## Pull Request Checklist

## Please don't

### DIFF
--- a/include/fkYAML/detail/input/lexical_analyzer.hpp
+++ b/include/fkYAML/detail/input/lexical_analyzer.hpp
@@ -823,17 +823,33 @@ private:
                     // Allow a space in an unquoted string only if the space is surrounded by non-space characters.
                     // See https://yaml.org/spec/1.2.2/#733-plain-style for more details.
                     current = m_input_handler.get_next();
+
+                    // These characters are permitted when not inside a flow collection, and not inside an implicit key.
+                    // TODO: Support detection of implicit key context for this check.
+                    if (m_flow_context_depth > 0)
+                    {
+                        switch (current)
+                        {
+                        case '{':
+                        case '}':
+                        case '[':
+                        case ']':
+                        case ',':
+                            return lexical_token_t::STRING_VALUE;
+                        }
+                    }
+
                     switch (current)
                     {
                     case ' ':
                     case '\r':
                     case '\n':
-                    case '{':
-                    case '}':
-                    case '[':
-                    case ']':
-                    case ',':
                     case ':':
+                        // " :" is permitted in a plain style string token, but not when followed by a space.
+                        if (!m_input_handler.test_next_char(' '))
+                        {
+                            break;
+                        }
                     case '#':
                     case '\\':
                         return lexical_token_t::STRING_VALUE;
--- a/include/fkYAML/detail/input/lexical_analyzer.hpp
+++ b/include/fkYAML/detail/input/lexical_analyzer.hpp
@@ -841,15 +841,15 @@ private:
 
                     switch (current)
                     {
-                    case ' ':
-                    case '\r':
-                    case '\n':
                     case ':':
                         // " :" is permitted in a plain style string token, but not when followed by a space.
                         if (!m_input_handler.test_next_char(' '))
                         {
                             break;
                         }
+                    case ' ':
+                    case '\r':
+                    case '\n':
                     case '#':
                     case '\\':
                         return lexical_token_t::STRING_VALUE;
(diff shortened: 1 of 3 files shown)


stephenwhittle:
As you can see, there's a fair bit of noise on this PR from me tweaking the tests to ensure proper coverage, and forgetting to run the amalgamation script a time or two. Serves me right for working late ;) 
However, is that generating a bunch of noise on your end for the failed runs? 
If so, is there a way I can import your Actions workflows into my fork? Would rather do all this iteration over there prior to opening a PR, instead of spamming you with notifications about failures.

fktn-k:
@stephenwhittle   
First of all, no worries!  
It didn't bother me since I've turned off the notification before when I did the same LOL.  

Rather, I'm sorry for not mentioning the way of checking test coverage in the CONTRIBUTING.md, 
I usually develop this library and run `lcov` & `genhtml` (as done in the Makefile [here](https://github.com/fktn-k/fkYAML/blob/develop/Makefile#L156)) on Ubuntu 22.04LTS (WSL2).  
I use the default version fetched from the apt repository.  
If you don't use WSL, though not for the "pure" Windows environment, I've seen some article which said that the tools are also available with pacman on MSYS2.  
Hope this will help.  

> is there a way I can import your Actions workflows into my fork?

Regarding this one, I honestly don't know at all...  
I'll look for some ways to achieve that for future contributions though.

fktn-k:
@stephenwhittle   
I've found a way in which you can run the actions in your fork repository after some experiments in my fork from another one.  

1. Go to the "Settings >> Actions >> General" settings section
2. Make sure the "Actions permissions" setting in your fork repository is not set to "Disable actions".  
3. Update the develop branch in your fork repository up-to-date
4. Create a PR to YOUR develop branch to check the workflow results.  
   When your changes get ready, just discard the PR if it's not necessary anymore.  
6. Create a PR to my develop branch as usual.

I've confirmed that the above steps do nothing to the upstream repository.  
If they wouldn't work to this repository... I don't care! Just give it a try!

stephenwhittle:
A few notes @fktn-k -
* Coveralls gives me a clickable link to the build, but when trying to inspect a coverage report for a specific file I get HTTP 500 errors from their site. It would be worth considering an improvement to the code coverage Github action, so that the coverage data for a failed check is dumped into the action's log or otherwise made available for the contributor to access in some way. I eventually got coverage information generated using WSL2 as per your suggestion, but for a one-off contributor an alternative way to view a failed report would be a usability improvement.

* The coverage issue was caused by `test_next_char` checking the current character to detect the end of input, which of course will never be true in the context of the code in which I was using it. The end-of-input check is being tested elsewhere in the codebase, but I suspect `test_next_char` was being inlined at my call-site, causing the coverage tests to want me to explicitly test the end-of-input check myself. 
  * It might be worth refactoring or restructuring such a function so it can be used more generally, as it's less prone to errors in the case of `get_next` (ie no risk of a forgotten call to `unget`) when we only need a single-character lookahead. 
  * It may also be worth investigating if there are other compiler flags we should be passing to disable inlining during code coverage checks as it seems to be occurring even at O0 optimization level.

fktn-k:
@stephenwhittle   
Thanks for sharing the notes!  

> Coveralls gives me a clickable link to the build, but when trying to inspect a coverage report for a specific file I get HTTP 500 errors from their site.

This happens to me as well... (I got a 504 (timeout) error. The same on your end?)  
image  
The same error happens to other repositories (see [this issue](https://github.com/lemurheavy/coveralls-public/issues/1751)), and there seems no way to get a report back from Coveralls (see [this comment](https://github.com/lemurheavy/coveralls-public/issues/1735#issuecomment-1826431737)). So, I'm looking to create a CI step to upload a zipped HTML view generated from the uploaded coverage data, what do you think?  

> It might be worth refactoring or restructuring such a function so it can be used more generally, as it's less prone to errors in the case of `get_next` (ie no risk of a forgotten call to `unget`) when we only need a single-character lookahead.  

That's more reasonable. I agree with the change to make it less prone to errors, like replacing `test_next_char` with something like `peek_next` which just returns a next character without changing the current position.  

> It may also be worth investigating if there are other compiler flags we should be passing to disable inlining during code coverage checks as it seems to be occurring even at O0 optimization level.  

Agree. I made some workarounds to avoid such issues before (I just recalled that). There should be some option to keep the compiler from expanding any functions inline.

fktn-k (approved):
Looks good to me. Thanks a lot, @stephenwhittle.

Issue #378, opened 2024-08-19
Infinity loop when trying to deserialize non existing file


sndth:
### Description

When you try to load a file that doesn't exist (using `std::ifstream`), `InputType` will be null and from what I saw, this isn't checked anywhere. I suppose this should be checked inside `input_adapter` function:

https://github.com/fktn-k/fkYAML/blob/93b0e43abecbfb8025aeeb835eba57358c604f47/include/fkYAML/detail/input/input_adapter.hpp#L926-L932

this issue doesn't occur when `std::FILE*` is used, because:

https://github.com/fktn-k/fkYAML/blob/93b0e43abecbfb8025aeeb835eba57358c604f47/include/fkYAML/detail/input/input_adapter.hpp#L915-L924

we have condition to check file pointer inside `input_adapter` for `std::FILE*`

### Reproduction steps

1. Use code from tutorial (https://fktn-k.github.io/fkYAML/tutorials/#__tabbed_1_2)
2. Make sure file doesn't exist
3. Compile and test = infinity loop because stream is empty

### Expected vs. actual results

Expected result is throw fkyaml::exception, currently we have infinity loop

### Compiler and operating system

Windows: C++latest, MSVC latest

### Library version

0.3.9/0.3.10

### Validation

- [X] The bug also occurs if the latest version from the [`develop`](https://github.com/fktn-k/fkYAML/tree/develop) branch is used.
- [ ] I can successfully [compile and run the unit tests](https://github.com/fktn-k/fkYAML#how-to-execute-the-unit-tests).

fktn-k:
@sndth   
Thanks for filing the issue.  
You're right.  
The input_adapter for `std::istream` should behave like that for `FILE*` if a file isn't opened by a library user, i.e., throw an `fkyaml::exception`.  
That's much much better and reasonable as an error handling behavior as well.  

By the way, do you want to fix it (because you've already found a way to resolve it)?  
Or I can do it tonight.  
Either is fine to me.

sndth:
I prepared PR which fixes the problem

fktn-k:
@sndth   
Thanks for the fix!  
I've merged the PR since everything your work looks fine (some unrelated CI issues happened though).

PR #379, opened 2024-08-20
Throw exception when loading a nonexistent stream

sndth:
This PR fixes the issue with attempting to call a deserialization function on a nonexistent stream (std::ifstream), reported in issue #378.

Previously, when attempting to read data from a nonexistent stream, an infinite loop would occur, trying to read data that was not there.

Now, if the stream is empty, an exception will be thrown, which can be handled using a try-catch block.

---

## Pull Request Checklist

## Please don't


fktn-k (approved):
Thanks for the fix!  
The changes look good to me.

fktn-k:
@sndth   
Although some CI jobs didn't pass successfully, that is not due to the changes introduced in this PR but some token accessibility issue in GA workflows. Just in case, I manually checked that the codebase is well formatted and properly amalgamated with no decrease in the coverage rate.  

So I'll merge this PR. Thanks a lot!

PR #391, opened 2024-09-07
Improve conversion from scalars to native types

fktn-k:
In this PR, conversions from scalars to native data types (int or bool) have been reimplemented, which happens when each scalar token is found in an input buffer during the deserialization process.  
The conversions required unnecessary std::string instance creations before, which causes a number of heap allocations even when the instanciations is just the same as duplication of some part of an input buffer.  
Most of the unnecessary heap allocations have therefore been replaced with references using iterators with no changes in public APIs and behaviors from the viewpoint of library users.  

One exception is that conversions from a scalar to a floating point value still requires unnecessary heap allocations by default.  
If `std::from_chars()` (since C++17) is available, the library now call it and thus no heap allocation happens.  
In the other cases, however, the library calls `std::stof()` or `std::stod()` depending on the destination type which require a `std::string` object for a null-terminated string.  
Although using some third party libraries like fast_float or implementing a similar feature inside the library can resolve this issue, that either consts a lot or leads to other issues (licensing, dependency management, etc...).  
So, I put aside the issue and moved the latest implementations (with some sophistication though).  

---

## Pull Request Checklist

## Please don't


fktn-k:
## CAVEAT

While fixing CI errors, I've found some issues related to Valgrind and MinGW compilers.  
The issues are all related to the `long double` type which were not well tested before.  

* ( **Valgrind** )  
  It doesn't support `long double` and its internal workaround somehow causes wrong errors.  
* ( **MinGW compilers** )  
  It produces weired conversion results from scalars(string) to long double values.  

All the related test cases are now changed to be disabled when either Valgrind or MinGW is used.

Issue #429, opened 2024-11-17
node.hpp(1351): Error C4800 : Implicit conversion from 'int' to bool. Possible information loss


tomwpid:
### Description

The following warning is generated when compiling with the node.hpp header included.

### Reproduction steps

Include node.hpp without the warning C4800 disabled.

### Expected vs. actual results

Expected: No warning should be generated.
Actual: Warning C4800 is generated.

### Minimal code example

```Shell
default:
            // alphabets and numbers are also allowed.
            return std::isalnum(c);
        }
```

### Error messages

```Shell
node.hpp(1351): Error C4800 : Implicit conversion from 'int' to bool. Possible information loss
```

### Compiler and operating system

Compiler: MSVC v143 - v14.38 - 17.8 OS: Windows

### Library version

0.3.14

### Validation

- [ ] The bug also occurs if the latest version from the [`develop`](https://github.com/fktn-k/fkYAML/tree/develop) branch is used.
- [ ] I can successfully [compile and run the unit tests](https://github.com/fktn-k/fkYAML#how-to-execute-the-unit-tests).

fktn-k:
@tomwpid   
Thanks for filing this issue.  
The warning C4800 is not enabled with the compile option `/W4` and I didn't know the warning is raised.  
I could reproduce the warning by passing the compile option `/w44800` in when building the test suite.  
I'll resolve the warning soon (I'm working on it), but I'm not planning to pass the compile `/Wall` since it enables so many warnings which I wouldn't like to take care of...

tomwpid:
OK - thank you!

fktn-k:
@tomwpid   
The C4800 warning should be resolved by the PR #430.  
Could you confirm the fix with the develop HEAD?

tomwpid:
@fktn-k - Confirmed, PR https://github.com/fktn-k/fkYAML/pull/430 has fixed it.

fktn-k:
@tomwpid   
Thanks! I'll close this issue then.  
Feel free to reopen it if you find anything related but unresolved.

PR #430, opened 2024-11-18
Resolve the C4800 warning when compiled with MSVC

fktn-k:
This PR fixes the [C4800](https://learn.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-3-c4800?view=msvc-170) warnings when this library is compiled with a MSVC compiler as reported in the issue #429.  
The warnings are now resolved by `static_cast`ing the result of `std::isalnum()` as `bool`.  
Furthermore, the test suite is now built with the compile option `/w44800` to avoid a possible regression in the future (the C4800 warning is desabled by default and even with the compile option `/W4`).  

---

## Pull Request Checklist

## Please don't

### DIFF
--- a/include/fkYAML/detail/encodings/uri_encoding.hpp
+++ b/include/fkYAML/detail/encodings/uri_encoding.hpp
@@ -115,7 +115,7 @@ private:
             return true;
         default:
             // alphabets and numbers are also allowed.
-            return std::isalnum(c);
+            return static_cast<bool>(std::isalnum(c));
         }
     }
 };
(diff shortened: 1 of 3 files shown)


fktn-k:
macos-12 is now deprecated and it's during the scheduled brownout period now.  
https://github.com/actions/runner-images/issues/10721  
That's why all the workflow jobs using the macos-12 runner are failing...  
I'll re-run those failed jobs tomorrow.

Issue #428, opened 2024-11-17
Valid YAML file triggers throw parse_error("Failed to convert a scalar to a floating point value", m_line, m_indent);


tomwpid:
### Description

Trying to load a large but valid YAML file results in the following exception.

throw parse_error("Failed to convert a scalar to a floating point value", m_line, m_indent);

### Reproduction steps

1. Download the problem file [Problem YAML.zip](https://github.com/user-attachments/files/17792281/Problem.YAML.zip)
2. Try to deserialise the file as below.
3. Observe the exception being thrown

### Expected vs. actual results

Expected: YAML should load without error
Actual: YAML cannot be loaded

### Minimal code example

```Shell
// Open a YAML file. Other streams or strings are also usable as an input.
td::ifstream ifs("invNames.yaml");

// Deserialize the loaded file contents.
fkyaml::node RootNode = fkyaml::node::deserialize(ifs);
```

### Error messages

```Shell
throw parse_error("Failed to convert a scalar to a floating point value", m_line, m_indent);
```

### Compiler and operating system

Compiler: MSVC v143 - v14.38 - 17.8 OS: Windows

### Library version

0.3.14

### Validation

- [ ] The bug also occurs if the latest version from the [`develop`](https://github.com/fktn-k/fkYAML/tree/develop) branch is used.
- [ ] I can successfully [compile and run the unit tests](https://github.com/fktn-k/fkYAML#how-to-execute-the-unit-tests).

fktn-k:
@tomwpid   
Thanks for filing the issue!  
The error is reproduced in my local environment too.  
I then investigated the error and found 2 issues in parsing the YAML file.  
It should not take much to fix them, but a workaround for both of them is to single or double quote relevant scalars.  

1. `6E-578` on the line 32882 of invNames.yaml is interpretted as a floating point scalar, but not expressible as the type `double`  
   ```yaml
   # extracted from invNames.yaml (lines 32881-32882)
   - itemID: 30003270
   itemName: 6E-578 # the 6E-578 is interpretted as a floating point scalar but not expressible as the 
   type `double`
   ```

   This causes the error in the description.  
   Although the interpretation itself is correct as per the regular expression for floating point values defined in [the YAML specification](https://yaml.org/spec/1.2.2/#1022-tag-resolution), errors in the string-to-float conversion should be recovered by treating the scalar as a string scalar.  
   However, I'd still recommend the use of either single/double quotes to resolve ambiguity.
[...]

PR #432, opened 2024-11-24
Support parsing multiline plain scalars

fktn-k:
In this PR, parsing multiline plain (unquoted) scalars has been implemented, which was described as a restriction in the PR #344  
> **The flow line folding for plain scalars is not yet supported** since the deserialization feature is not yet ready for that feature. It will, however, surely be supported in the future. Until the feature gets ready, please make such a scalar be either single or double quoted.

Before this PR, multiline plain scalars seemed to be parsed with no error but the parse result was just incorrect since its data structure was different from the expected one. (I know this was terrible in debugging...)  
The library can now successfully parses such scalars and the result has its correct data structure.  
We can simply pass this YAML
```yaml
key: this line
  is a multiline
  plain scalar.
```

to the `fkyaml::node::deserialize()` function, and the function then interprets the input as if the following YAML is passed.
```yaml
key: this line is a multiline plain scalar.
```

Note that, although multiline scalars (either plain or single/double quoted) cannot be used as an implicit mapping key as stated in [the YAML specification (7.3. Flow Scalar Styles)](https://yaml.org/spec/1.2.2/#73-flow-scalar-styles), the library currently allows for such an invalid syntax like this:
```yaml
this implicit
  multiline
  scalar: should not be allowed
```

This will be fixed in another PR for sure.  

---

## Pull Request Checklist

## Please don't

### DIFF
--- a/include/fkYAML/detail/input/lexical_analyzer.hpp
+++ b/include/fkYAML/detail/input/lexical_analyzer.hpp
@@ -799,7 +799,7 @@ private:
                         // If so, the indent value remains the current one.
                         // Otherwise, the indent value is changed based on the last ocurrence of the above 3.
                         // In any case, multiline plain scalar content must be indented more than the indent value.
-                        str_view line_content_part {line_begin_itr + indent, &sv[pos]};
+                        const str_view line_content_part {line_begin_itr + indent, &sv[pos]};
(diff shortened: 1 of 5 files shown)


fktn-k:
This also fixes one if of the issues in parsing the attached YAML file to the issue #428.

PR #497, opened 2025-05-11
Fix project compilation on GCC compilers

sndth:
This PR fixes an issue with project compilation on GCC compilers.
Looks like the newer GCC versions shows `-Werror=deprecated-literal-operator` when the project is building.

I prepared a simple fix that check whether compiler is `__GNUC__` and the version is newer than `6`.
If yes, a new macro - `FK_YAML_QUOTE_OPERATOR` will use `operator""_yaml`, otherwise `operator"" _yaml`.

---

## Pull Request Checklist

## Please don't

### DIFF
--- a/include/fkYAML/node.hpp
+++ b/include/fkYAML/node.hpp
@@ -1951,12 +1951,18 @@ inline namespace yaml_literals {
 #pragma clang diagnostic ignored "-Wdeprecated"
 #endif
 
+#if defined(__GNUC__) && (__GNUC__ > 6)
+#define FK_YAML_QUOTE_OPERATOR operator""_yaml
+#else
+#define FK_YAML_QUOTE_OPERATOR operator"" _yaml
+#endif
+
 /// @brief The user-defined string literal which deserializes a `char` array into a `node` object.
 /// @param s An input `char` array.
 /// @param n The size of `s`.
 /// @return The resulting `node` object deserialized from `s`.
 /// @sa https://fktn-k.github.io/fkYAML/api/operator_literal_yaml/
-inline fkyaml::node operator"" _yaml(const char* s, std::size_t n) {
+inline fkyaml::node FK_YAML_QUOTE_OPERATOR(const char* s, std::size_t n) {
     return fkyaml::node::deserialize(s, s + n);
 }
 
@@ -1965,7 +1971,7 @@ inline fkyaml::node operator"" _yaml(const char* s, std::size_t n) {
 /// @param n The size of `s`.
 /// @return The resulting `node` object deserialized from `s`.
 /// @sa https://fktn-k.github.io/fkYAML/api/operator_literal_yaml/
-inline fkyaml::node operator"" _yaml(const char16_t* s, std::size_t n) {
+inline fkyaml::node FK_YAML_QUOTE_OPERATOR(const char16_t* s, std::size_t n) {
     return fkyaml::node::deserialize(s, s + n);
 }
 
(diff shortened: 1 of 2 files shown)


fktn-k:
@sndth   
Thanks for dealing with this.  
Taking a brief look at the changes, I guess you're using different version of clang-format.  
Please use the script under scripts directory.

include/fkYAML/node.hpp
```diff
@@ -1951,12 +1951,18 @@ inline namespace yaml_literals {
 #pragma clang diagnostic ignored "-Wdeprecated"
 #endif
 
+#if defined(__GNUC__) && (__GNUC__ > 6)
```
fktn-k, line 1954: @sndth   
Maybe you can just replace __clang__ with __GNUC__ on line 1949?

sndth:
> @sndth Thanks for dealing with this. Taking a brief look at the changes, I guess you're using different version of clang-format. Please use the script under scripts directory.

Oh exactly, I am using 20.1.4 but the project is using 18.1.3.
If you want, I can bump version of clang-format, because changes between these versions in case of this project are minimal.

include/fkYAML/node.hpp
sndth: > @sndth Maybe you can just replace **clang** with **GNUC** on line 1949?

We will then have GCC compiler errors about `#pragma clang`.

I tried to use something like this:
```cpp
#if defined(__GNUC__)
#pragma GCC diagnostic push
#pragma GCC diagnostic ignored "-Wdeprecated-declarations"
#elif defined(__clang__)
#pragma clang diagnostic push
#pragma clang diagnostic ignored "-Wdeprecated"
#endif
```

But it doesn't help, and `-Werror=deprecated-literal-operator` error is still showing.

fktn-k:
> > @sndth Thanks for dealing with this. Taking a brief look at the changes, I guess you're using different version of clang-format. Please use the script under scripts directory.
> 
> Oh exactly, I am using 20.1.4 but the project is using 18.1.3. If you want, I can bump version of clang-format, because changes between these versions in case of this project are minimal.

Clang-format tools with different major versions would cause a lot of unrelated changes. Although bumping-up is ok, I'd say it's better to do so in a different PR.  
Can you use `scripts/run_clang_format.(sh|bat)` which forces clang-format 18.1.3 for now?

include/fkYAML/node.hpp
fktn-k: The error occurs because Clang also defines `__GNUC__`.  
So, something like this should work? (not tried though)  
```cpp
#if defined(__GNUC__)
#if defined(__clang__)
#pragma clang diagnostic push
#pragma clang diagnostic ignored "-Wdeprecated"
#else // maybe GCC version check here?
#pragma GCC diagnostic push
#pragma GCC diagnostic ignored "-Wdeprecated-declarations"
#endif
#endif
```

sndth:
> > > @sndth Thanks for dealing with this. Taking a brief look at the changes, I guess you're using different version of clang-format. Please use the script under scripts directory.
> > 
> > 
> > Oh exactly, I am using 20.1.4 but the project is using 18.1.3. If you want, I can bump version of clang-format, because changes between these versions in case of this project are minimal.
> 
> Clang-format tools with different major versions would cause a lot of unrelated changes. Although bumping-up is ok, I'd say it's better to do so in a different PR. Can you use `scripts/run_clang_format.(sh|bat)` which forces clang-format 18.1.3 for now?

Yep! I just dropped commit with formatting changes and there is no files to reformat for now.

include/fkYAML/node.hpp
sndth: Even with GCC version check like:
```cpp
#if defined(__GNUC__)
  #if defined(__clang__)
    #pragma clang diagnostic push
    #pragma clang diagnostic ignored "-Wdeprecated"
  #else
  #if __GNUC__ > 6
    #pragma GCC diagnostic push
    #pragma GCC diagnostic ignored "-Wdeprecated-declarations"
  #endif
#endif
```
Problem still existing on some GCC versions.
Probably I am doing something wrong :/, so sorry!

If you have other idea please let me know, but from what I see, in e.g. [nlohmann::json](https://github.com/nlohmann/json/pull/4161/commits/6b8456116737733fc90122868ad55e6b22768b6d) library, similiar issue also was resolved the same way like in my commit.

fktn-k: @sndth  
OK, I finally agree with your idea.  
Thanks for your response.

fktn-k (approved):
Looks good to me. Thanks!

PR #508, opened 2025-11-17
Fix serialization of strings starting with YAML indicator characters

rioki:
This PR corrects a round-trip issue where certain string values were emitted as plain scalars even though YAML interprets them as structural tokens. For example, a value like `"***"` was serialized as:

```yaml
pass: ***
```

which YAML parses as an alias, causing deserialization to fail.

The serializer now performs a conservative scan of escaped string contents and quotes strings that begin with, or contain, YAML indicator characters that could change parsing semantics (e.g. `*`, `&`, `!`, `#`, `[`/`]`, `{`/`}`, etc.). This preserves the intended string type without modifying the parser or `scalar_scanner`.

All existing tests pass, and the added round-trip test covering this case succeeds.

---

## Pull Request Checklist

## Please don't

### DIFF
--- a/include/fkYAML/detail/output/serializer.hpp
+++ b/include/fkYAML/detail/output/serializer.hpp
@@ -380,38 +380,42 @@ private:
         }
 
         for (char c : str) {
-            switch (c){
-                // Definitely dangerous as plain: will be lexed as alias/anchor/tag/comment
-                case '*': // alias: *foo
-                case '&': // anchor: &foo
-                case '!': // tag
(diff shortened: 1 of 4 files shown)


rioki:
Not sure why the CI tripped... it failed on git checkout... If I can do anything to help on my end.

fktn-k:
@rioki 
Sorry for my response being too late.
Also, thanks a lot for your contribution.
I will close this PR in favor of #523 which is also aimed to fix the same issue.

PR #523, opened 2026-07-06
Fix serialization of strings that are invalid plain scalars

### LINKED ISSUE #517: Serializing strings with special chars
### Description

Serializing nodes with string values that start with special yaml syntax chars (see https://www.tutorialspoint.com/yaml/yaml_syntax_characters.htm) results in an invalid yaml string.
These values should have quotes around them.
Other tools/libs are throwing a syntax error when trying to read the output of fkyaml.

### Reproduction steps

```c++
std::string yaml = R"(
project: fkYAML
percent: %
startingChar: %test
)";
auto node = fkyaml::node::deserialize(yaml);

std::ofstream ofs("out.yaml");
ofs << node;
```

Expected:
```yaml
percent: "%"
project: fkYAML
startingChar: "%test"
```

actual:
```yaml
percent: %
project: fkYAML
startingChar: %test
```

### Minimal code example

```Shell

```

### Error messages

```Shell

```

Win11 MSVC 20

### Library version

0.4.2

### Validation

- [ ] The bug also occurs if the latest version from the [`develop`](https://github.com/fktn-k/fkYAML/tree/develop) branch is used.
- [ ] I can successfully [compile and run the unit tests](https://github.com/fktn-k/fkYAML#how-to-execute-the-unit-tests).

sndth:
This PR fixes serialization of string scalars that cannot be safely emitted in YAML plain style.

Previously, the serializer only added quotes when emitting a string plainly would change its scalar type, for example `"123"` becoming an integer. That missed strings that still resolve as strings but are invalid YAML syntax when emitted without quotes, such as values starting with directive, reserved, or flow indicator characters.

This PR updates string serialization to quote those unsafe plain scalars, including values starting with characters such as `%`, `@`, `!`, `&`, and `{`.

It also adds regression coverage for the reported `%` / `%test` mapping case and related plain-scalar indicator cases.

Fixes #517 

---

## Pull Request Checklist

## Please don't

### DIFF
--- a/include/fkYAML/detail/output/serializer.hpp
+++ b/include/fkYAML/detail/output/serializer.hpp
@@ -384,7 +384,7 @@ private:
 
         typedef typename BasicNodeType::string_type string_type;
         const typename string_type::size_type npos = string_type::npos;
-        if (s.find(": ") != npos || s.find(":\t") != npos || s.find(" #") != npos || s.find("\t#") != npos) {
+        if (s.find(": ") != npos || s.find(" #") != npos) {
             return false;
         }
 
@@ -392,7 +392,7 @@ private:
         case '-':
         case '?':
         case ':':
-            return s.size() > 1 && s[1] != ' ' && s[1] != '\t';
+            return s.size() > 1 && s[1] != ' ';
         case ',':
         case '[':
         case ']':
@@ -410,10 +410,9 @@ private:
         case '@':
         case '`':
         case ' ':
-        case '\t':
             return false;
         default:
-            return s.back() != ' ' && s.back() != '\t';
+            return s.back() != ' ';
         }
     }
 
--- a/include/fkYAML/detail/output/serializer.hpp
+++ b/include/fkYAML/detail/output/serializer.hpp
@@ -383,8 +383,7 @@ private:
         }
 
         typedef typename BasicNodeType::string_type string_type;
-        const typename string_type::size_type npos = string_type::npos;
-        if (s.find(": ") != npos || s.find(" #") != npos) {
+        if (s.find_first_of(" \t\n\r,[]{}") != string_type::npos) {
             return false;
         }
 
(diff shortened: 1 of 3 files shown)


include/fkYAML/detail/output/serializer.hpp
```diff
@@ -377,6 +377,45 @@ class basic_serializer {
...
+        switch (s.front()) {
+        case '-':
+        case '?':
+        case ':':
+            return s.size() > 1 && s[1] != ' ';
+        case ',':
+        case '[':
+        case ']':
+        case '{':
+        case '}':
```
fktn-k: Given that this function doesn't know whether this plain scalar is an implicit key or inside a flow collection, `,`, `[`, `]`, `{` and `}` should be checked against all the characters in a string.
For simplicity, I'm OK with always wrapping a string which contains any of those characters with double quotes.

include/fkYAML/detail/output/serializer.hpp
```diff
@@ -377,6 +377,45 @@ class basic_serializer {
...
+        case '*':
+        case '!':
+        case '|':
+        case '>':
+        case '\'':
+        case '"':
+        case '%':
+        case '@':
+        case '`':
+        case ' ':
```
fktn-k, line 406: A white space is also allowed in the middle of a string only if it's followed by a non-space character.
That said, how about always wrapping a string which contains any white space with double quotes for simplicity?

fktn-k (changes requested):
@sndth 
Thank you for the contribution!
I've reviewed your changes and posted some comments.

include/fkYAML/detail/output/serializer.hpp
sndth: Thanks, that makes sense. The serializer does not know whether the scalar is used as an implicit key or inside a flow collection, so a conservative rule is better here.

I updated the helper to quote strings containing any whitespace or any of `,[]{}` regardless of position. I also adjusted the tests so `foo bar` is quoted and added coverage for flow indicators appearing in the middle of a string.

fktn-k (approved):
@sndth 
Looks good to me. Thanks!

Issue #502, opened 2025-07-30
Defect in c-tor


KristianIvarsson:
### Description

There seem to  be some flaw in move c-tor that makes the type not become as explected

Maybe a simple `std::variant` and using default implementation would be more simple

### Reproduction steps

```
   const fkyaml::node a = fkyaml::node::mapping();
   const fkyaml::node b{ fkyaml::node::mapping()};
   assert( a.get_type() == b.get_type());
```

### Expected vs. actual results

Expected
```
assert( b.get_type() == fkyaml::node::MAPPING);
```
Actual
```
assert( b.get_type() == fkyaml::node::SEQUENCE);
```

### Minimal code example

```Shell

```

### Error messages

```Shell

```

### Compiler and operating system

g++13 Ubuntu 

### Library version

g++13

### Validation

- [ ] The bug also occurs if the latest version from the [`develop`](https://github.com/fktn-k/fkYAML/tree/develop) branch is used.
- [ ] I can successfully [compile and run the unit tests](https://github.com/fktn-k/fkYAML#how-to-execute-the-unit-tests).

sndth:
Hi! Thank you for reporting!
This (probably) isn't a bug in the library, but a result of how different compilers handle move semantics:

```cpp
const fkyaml::node b{ fkyaml::node::mapping()};
```
`GCC` and `Clang` perform a move from the temporary returned by `mapping()`, which clears internal state (like `m_attrs`) in the move constructor. `MSVC`, on the other hand, aggressively applies copy elision, so no move happens - the temporary is constructed directly into `b`, preserving the type.

That's why `get_type()` returns different results on each compiler.
Best fix is avoid brace initialization with temporaries. Use assignment instead:
```cpp
const auto b = fkyaml::node::mapping(); // Consistent across all compilers
```
Let me know if you'd prefer a patch that preserves type info after move!

KristianIvarsson:
Yes, I guess RVO is applied in the former construct, but the observable behaviour should be the same in the latter case

I'm not sure if users always have the possibility to avoid brace initialization (e.g. a  c-tor for a const member or such)

>Let me know if you'd prefer a patch that preserves type info after move!

I'm not 100% sure what you mean by _preserve_, but isn't that what should happen? Shouldn't every property of _rhs_ be moved/copied into _lhs_?

KristianIvarsson:
I looked a bit more into this @sndth 

[...]

PR #524, opened 2026-07-06
Fix brace initialization of empty collection nodes (issue #502)

sndth:
This PR introduces a narrow fix for brace initialization of empty collection nodes.

When code uses:

```cpp
const fkyaml::node node{fkyaml::node::mapping()};
```

the initializer-list constructor is selected, so the empty mapping node was previously treated as a single sequence element. As a result, the constructed node became a sequence instead of preserving the mapping type.

This change tracks whether an initializer-list entry was created directly from a `basic_node`, then unwraps only the narrow case where the list contains exactly one empty mapping or empty sequence node without node properties.

The fix intentionally keeps existing scalar initializer-list behavior unchanged. For example, `node{"test"}` still creates a one-element sequence, as documented.

Added regression coverage for:
- `node{node::mapping()}` preserving mapping type
- `node{node::sequence()}` preserving sequence type
- `node{"test"}` still creating a one-element sequence

---

## Pull Request Checklist

## Please don't

### DIFF
--- a/include/fkYAML/detail/node_ref_storage.hpp
+++ b/include/fkYAML/detail/node_ref_storage.hpp
@@ -32,13 +32,15 @@ public:
     /// @brief Construct a new node ref storage object with an rvalue basic_node object.
     /// @param n An rvalue basic_node object.
     explicit node_ref_storage(node_type&& n) noexcept(std::is_nothrow_move_constructible<node_type>::value)
-        : m_owned_value(std::move(n)) {
+        : m_owned_value(std::move(n)),
+          m_has_node_ref(true) {
     }
 
     /// @brief Construct a new node ref storage object with an lvalue basic_node object.
     /// @param n An lvalue basic_node object.
     explicit node_ref_storage(const node_type& n) noexcept
-        : m_value_ref(&n) {
+        : m_value_ref(&n),
+          m_has_node_ref(true) {
     }
 
     /// @brief Construct a new node ref storage object with a std::initializer_list object.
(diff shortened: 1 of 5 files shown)


include/fkYAML/detail/node_ref_storage.hpp
```diff
@@ -31,14 +31,16 @@ class node_ref_storage {
 public:
     /// @brief Construct a new node ref storage object with an rvalue basic_node object.
     /// @param n An rvalue basic_node object.
-    explicit node_ref_storage(node_type&& n) noexcept(std::is_nothrow_move_constructible<node_type>::value)
-        : m_owned_value(std::move(n)) {
+    node_ref_storage(node_type&& n) noexcept(std::is_nothrow_move_constructible<node_type>::value)
```
fktn-k: Why is the `explicit` no lonher necessary? IIRC, it was intended for unnecesary implicit conversions to not occur.

fktn-k (commented):
@sndth 
Thanks for yourcontribution.
I have reviewed the changes.
It would be nice if the documentation in the docs directory is updated as well.

include/fkYAML/detail/node_ref_storage.hpp
sndth: Thanks for the review.

I removed `explicit` because GCC/Clang were selecting a different constructor path for this initializer-list case than MSVC, which made the fix pass locally but fail across the CI matrix.

That said, I agree with your point that `explicit` should stay there to avoid unnecessary implicit conversions. I’ll try to handle this in the variadic constructor path instead, only for the single direct `basic_node` case, and push an update ASAP.

sndth: I restored `explicit` on the direct `basic_node` constructors.

The GCC/Clang issue was handled in the variadic constructor path instead. When that path receives exactly one `basic_node` argument, it now marks the storage as a direct node reference, so the initializer-list constructor can still preserve `node{node::mapping()}` and `node{node::sequence()}` as empty collection nodes.

I also updated the constructor documentation in `docs` 😄

fktn-k (approved):
@sndth 
Thanks for the changes.
Looks good to me now.
I will merge this into develop.

Issue #514, opened 2026-03-05
Assertion failure FK_YAML_ASSERT(!m_context_stack.empty()) when deserializing malformed YAML with empty key and invalid scalar


chemoontheshy:
### Description

When calling YamlNode::deserialize() with a specific malformed YAML string, an internal assertion FK_YAML_ASSERT(!m_context_stack.empty()) is triggered, causing the program to abort. The assertion cannot be caught using a try-catch block, as it is not a C++ exception but an assertion failure. This makes it impossible to handle the error gracefully in application code.

### Expected vs. actual results

I cannot prevent users from using incorrect YAML formats, so I need to validate the input first, and the program should not crash directly.

### Minimal code example

```Shell
    try {
        std::stringstream buffer;
        buffer << "\"\":\n  port:not_a_scalar]\n";  // malformed YAML input
        const std::string yaml_text = buffer.str();
        const YamlNode root = YamlNode::deserialize(yaml_text);
    }
    catch (...) {
        // This block is never entered because the failure is an assertion.
    }
```

### Error messages

```Shell
Assertion failed: !m_context_stack.empty(), file ..//fkYAML/detail/input/deserializer.hpp, line 464
```

### Compiler and operating system

 window11 gcc-15.2.0-mingw-w64ucrt-13.0.0-r5

### Library version

version 0.4.2

### Validation

- [ ] The bug also occurs if the latest version from the [`develop`](https://github.com/fktn-k/fkYAML/tree/develop) branch is used.
- [x] I can successfully [compile and run the unit tests](https://github.com/fktn-k/fkYAML#how-to-execute-the-unit-tests).

sndth:
Thanks for the report and sorry for late response
I tried to reproduce the exact input from the issue on the current `develop`, and also on `v0.4.2`:

```cpp
"\"\":\n  port:not_a_scalar]\n"
```

On my local environment, this exact input does not trigger the assertion. It is parsed as a mapping with an empty string key and the value `port:not_a_scalar]`. This also seems consistent with the current parser behavior, since flow indicators such as `]` are already accepted inside block plain scalars outside a flow context.

However, I found a closely related case which does reproduce the same assertion:

```yaml
: foo
```

This hits:

```text
FK_YAML_ASSERT(!m_context_stack.empty())
```

in the `KEY_SEPARATOR` handling path.

[...]

PR #525, opened 2026-07-06
Handle empty mapping key separator as parse error

sndth:
This PR introduces a defensive fix for malformed mapping key separators.

The deserializer previously relied on `FK_YAML_ASSERT(!m_context_stack.empty())` when handling a `KEY_SEPARATOR` token. For malformed input such as:

```yaml
: foo
```

the context stack can be empty, causing the process to abort instead of throwing a catchable exception.

This change replaces that assertion-only guard with a `parse_error` when a key separator is encountered without an active parsing context. This keeps malformed input on the normal error handling path and lets callers catch the failure.

I could not reproduce the exact assertion failure from issue #514 locally with:

```cpp
"\"\":\n  port:not_a_scalar]\n"
```

On my environment, that input is parsed as a mapping with an empty string key and a plain scalar value. I still added regression coverage for it to ensure deserialization does not abort.

Added regression coverage for:
- empty implicit key input (`: foo`) throwing `parse_error`
- the issue #514 input not aborting during deserialization

---

## Pull Request Checklist

## Please don't

### DIFF
--- a/include/fkYAML/detail/input/deserializer.hpp
+++ b/include/fkYAML/detail/input/deserializer.hpp
@@ -239,6 +239,17 @@ private:
             indent = lexer.get_last_token_begin_pos();
             break;
         }
+        case lexical_token_t::KEY_SEPARATOR:
+            root = basic_node_type::mapping();
+            apply_directive_set(root);
+            apply_node_properties(root);
+            m_context_stack.emplace_back(
+                lexer.get_lines_processed(), lexer.get_last_token_begin_pos(), context_state_t::BLOCK_MAPPING, &root);
+            add_new_key(basic_node_type(""), line, indent);
+            token = lexer.get_next_token();
+            line = lexer.get_lines_processed();
+            indent = lexer.get_last_token_begin_pos();
+            break;
         case lexical_token_t::BLOCK_LITERAL_SCALAR:
         case lexical_token_t::BLOCK_FOLDED_SCALAR:
             // If a block scalar token is detected here, current document contains single scalar.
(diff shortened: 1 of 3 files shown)


include/fkYAML/detail/input/deserializer.hpp
```diff
@@ -461,7 +461,9 @@ class basic_deserializer {
                 continue;
             }
             case lexical_token_t::KEY_SEPARATOR: {
-                FK_YAML_ASSERT(!m_context_stack.empty());
+                if FK_YAML_UNLIKELY (m_context_stack.empty()) {
+                    throw parse_error("mapping key should not be empty.", line, indent);
```
fktn-k: IIRC, an empty mapping key is allowed in the specification.
If possible cases where such a key appears in a YAML document are covered in this if branch, insert an empty string scalar as a mapping key to the node tree instead of throwing a parse_error. Otherwise, treat it as unsupported for now.

fktn-k (changes requested):

include/fkYAML/detail/input/deserializer.hpp
sndth: Thanks, good catch! I overlooked that empty mapping keys are valid YAML.

Updated the handling for the root-level case: `: foo` is now parsed as a mapping with an empty string key (`"" -> "foo"`) instead of throwing a `parse_error`.

The sequence cases such as `- : foo` and `- - : foo` are still treated as unsupported for now and continue to throw.

fktn-k: Thanks!
Let me make just one more request.
Could you add a test case to make sure that a parse_error is thrown in the sequence cases?

sndth: Thanks, added/clarified the regression coverage for those unsupported sequence cases.

The `Deserializer_KeySeparator` test now explicitly checks that both `- : foo` and `- - : foo` throw `parse_error`, while the root-level `: foo` case is still parsed as a mapping with an empty string key.

fktn-k (approved):
@sndth 
Looks good to me. Thanks!

PR #527, opened 2026-07-09
Fix `tool` directory typo and add CMake policy for the benchmark build

sndth:
This PR fixes a minor typo in the `tool` directory name (actually it should be `tools`) and resolves a build issue with `yaml-cpp` caused by an outdated CMake version.

---

## Pull Request Checklist

## Please don't

### DIFF
--- a/tools/benchmark/README.md
+++ b/tools/benchmark/README.md
@@ -29,7 +29,7 @@ Build and run this tool with the following commands:
 
 ```bash
 $ cd path/to/fkYAML
-$ cmake -S . -B build -DCMAKE_BUILD_TYPE={Debug|Release} -DFK_YAML_RUN_BENCHMARK=ON
+$ cmake -S . -B build -DCMAKE_BUILD_TYPE={Debug|Release} -DFK_YAML_RUN_BENCHMARK=ON -DCMAKE_POLICY_VERSION_MINIMUM=3.5
 $ cmake --build build --config {Debug|Release}
 
 # You can specify an arbitrary input file such as /foo/bar/sample.yml or /foo/bar/sample.json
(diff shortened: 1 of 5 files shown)


tools/benchmark/README.md
```diff
@@ -29,11 +29,11 @@ Build and run this tool with the following commands:
 
 ```bash
 $ cd path/to/fkYAML
-$ cmake -S . -B build -DCMAKE_BUILD_TYPE={Debug|Release} -DFK_YAML_RUN_BENCHMARK=ON
+$ cmake -S . -B build -DCMAKE_BUILD_TYPE={Debug|Release} -DFK_YAML_RUN_BENCHMARK=ON -DCMAKE_POLICY_VERSION_MINIMUM=3.5
```
fktn-k: If this definition is always required, it seems that this should be defined in the root CMakeLists.txt when FK_YAML_RUN_BENCHMARK is ON.

fktn-k (changes requested):
@sndth 
Thanks for your contribution.
I reviewed the changes.

tools/benchmark/README.md
sndth: Good point. I moved this into the root `CMakeLists.txt` so it's set automatically when `FK_YAML_RUN_BENCHMARK` is ON, before adding the benchmark subdirectory. I also removed the manual (`DCMAKE_POLICY_VERSION_MINIMUM=3.5` flag from the benchmark README).

Verified that configuring with `FK_YAML_RUN_BENCHMARK=ON` works without passing the policy variable manually.

fktn-k (approved):
@sndth 
Looks good to me. Thanks!

PR #507, opened 2025-10-20
added node::remove method

flaviu22:
[Describe your pull request here. Please read the description below and make sure you follow the checklist.]  

Added `basic_node::remove` method that remove a specific node. Tested, ran ok.

## Pull Request Checklist

Read the [CONTRIBUTING.md](https://github.com/fktn-k/fkYAML/blob/develop/.github/CONTRIBUTING.md) file for detailed information.  

- [ ] Changes are described in the pull request or in a referenced [issue](https://github.com/fktn-k/fkYAML/issues).
- [x] The test suite compiles and runs without any error.
- [x] [The code coverage](https://coveralls.io/github/fktn-k/fkYAML) on your branch is 100%.
- [ ] The documentation is updated if you added/changed a feature.

## Please don't

- The C++11 support varies between different **compilers** and versions. Please note the [list of supported compilers](https://github.com/fktn-k/fkYAML/blob/develop/README.md#supported-compilers). Some compilers like GCC 4.7 (and earlier), Clang 3.3 (and earlier), or Microsoft Visual Studio 13.0 and earlier are known not to work due to missing or incomplete C++11 support. Please refrain from proposing changes that work around these compiler's limitations with `#ifdef`s or other means.
- Please refrain from proposing changes that would **break [YAML](https://yaml.org/) specifications**. If you propose a conformant extension of YAML to be supported by the library, please motivate this extension.
- Please do not open pull requests that address **multiple issues**.

PR #530, opened 2026-07-16
Add tests and documentation for `basic_node::erase`

sndth:
This is a follow-up to #507.

The `basic_node::erase()` implementation was originally contributed by @flaviu22. This pull request completes the requested follow-up work by adding tests and documentation.

## Changes
- Add unit tests for:
  - erasing an existing key
  - erasing a missing key
  - using a `basic_node` as the key
  - erasing through an alias node
  - calling `erase()` on non-mapping nodes
- Add API documentation for `basic_node::erase()`
- Add a compilable documentation example
- Add the new API page to the documentation navigation
- Make the API name and error message consistent between the regular and single headers

## Testing
- All 370 unit tests pass.
- The new example compiles with both the regular and single-header versions.

Credit for the original `basic_node::erase()` implementation goes to @flaviu22. I only added the requested tests, documentation, and consistency fixes.

---

## Pull Request Checklist

## Please don't


include/fkYAML/node.hpp
```diff
@@ -1786,6 +1786,28 @@ class basic_node {
         return {*this};
     }
 
+    /// @brief Erase a mapping entry by key.
+    /// @tparam KeyType A type for the input key (any type convertible to node).
+    /// @param key A key identifying the mapping entry to erase.
+    /// @return true if an entry was found and erased, false otherwise.
+    template <typename KeyType>
+    bool erase(KeyType&& key) {
```
fktn-k: Upon reflection, I think it's better for this method to return the number of erased entries (0 or 1) in `size_type` to align the convention in STL like `std::map::erase`.
That would be more intuitive for library users and also helpful for us to add overloads with iterators in the future.

fktn-k (changes requested):
@sndth 
Thanks a lot for the follow-up.
I reviewed the changes and posted a comment.

include/fkYAML/node.hpp
sndth: Thank you! I updated `basic_node::erase()` to return `size_type` with either 0 or 1, matching the convention of `std::map::erase()`.

The tests now explicitly verify the return type and exact return values. All 370 tests pass.

fktn-k: Thank you for being parient with my request!

fktn-k (approved):
@sndth 
Looks good to me. Thanks!

Issue #492, opened 2025-03-25
encoded_buf_size underflow potential


LuanTechAutomation:
### Description

line 8880, in node.hpp (ver 0.4.2)
`
encoded_buf_size -= consumed_size;
`

Underflowed constant INTEGER_OVERFLOW potential when consumed_size > encoded_buf_size.
For example: encoded_buf_size = 1 and consumed_size > 1.

### Reproduction steps

when consumed_size > encoded_buf_size.
For example: encoded_buf_size = 1 and consumed_size > 1.

### Expected vs. actual results

The line encoded_buf_size -= consumed_size shouldn't cause underflow issue. Code need more secure.

### Minimal code example

```Shell

```

### Error messages

```Shell
Interge underflow.
```

### Compiler and operating system

g++, Linux

### Library version

0.4.2

### Validation

- [ ] The bug also occurs if the latest version from the [`develop`](https://github.com/fktn-k/fkYAML/tree/develop) branch is used.
- [ ] I can successfully [compile and run the unit tests](https://github.com/fktn-k/fkYAML#how-to-execute-the-unit-tests).

sndth:
Hi! Thanks for reporting this issue!
After quite a bit of debugging, I believe we have finally managed to resolve the problem 😄

I've opened PR #533, which probably fixes the potential underflow (so your case).
Also improves the handling of truncated encoded input, and adds regression tests for the reported case.

fktn-k:
Since #533 has been merged into develop, the reported potential issue should be resolved.
Please feel free to reopen it if any relavant issue is found.
Thank you, @sndth.

PR #533, opened 2026-07-19
Harden input adapter against truncated encoded input

sndth:
This PR hardens the input adapters against truncated UTF-8/16/32 input and fixes the potential `encoded_buf_size` underflow reported in #492.

Changes:
- Validate that all required bytes are available before decoding
- Prevent UTF-16 buffer-size underflow
- Clear consumed UTF-16 code units to avoid reusing stale data
- Reject incomplete input consistently across iterator, `FILE*`, and stream adapters
- Correctly handle BOM-only input and standalone UTF-16 carriage returns
- Safely create views for empty normalized buffers
- Add regression tests for truncated input, dangling UTF-16 surrogates, BOM-only input, and standalone carriage returns

All 367 unit tests pass and finally solves #492.

---

## Pull Request Checklist

## Please don't

### DIFF
--- a/include/fkYAML/detail/input/input_adapter.hpp
+++ b/include/fkYAML/detail/input/input_adapter.hpp
@@ -36,6 +36,17 @@ FK_YAML_DETAIL_NAMESPACE_BEGIN
 template <typename IterType, typename = void>
 class iterator_input_adapter;
 
+template <typename IterType>
+uint8_t read_utf8_byte_or_throw(IterType& current, IterType end, std::initializer_list<uint8_t> bytes_read) {
+    if FK_YAML_UNLIKELY (current == end) {
+        throw fkyaml::invalid_encoding("Invalid UTF-8 encoding.", bytes_read);
+    }
+
+    const auto byte = static_cast<uint8_t>(*current);
+    ++current;
+    return byte;
+}
+
 /// @brief An input adapter for iterators of type char.
 /// @tparam IterType An iterator type.
 template <typename IterType>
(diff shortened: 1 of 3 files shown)


sndth:
Hmm, coverage report was generated and uploaded successfully, but the workflow lost the final battle while trying to post a comment😆

Btw, looks like Coveralls is not entirely happy with this PR :/
The four uncovered lines are defensive checks for:
```
consumed_size > encoded_buf_size
```
After tracing the current UTF-16 buffer flow, these branches do not appear to be reachable through the public API. A one element buffer always has its second element cleared, so `from_utf16()` either consumes one element or throws before returning. The checks are still useful as protection against future regressions.

Would you prefer me to mark these defensive blocks with `LCOV_EXCL_START`/`LCOV_EXCL_STOP`, or handle them differently?

fktn-k:
@sndth 
Thank you for adressing the issue.
Let me take a moment to investigate the coverage issue.

fktn-k:
@sndth 
Sorry for my late response.
After some investigation and reflection, I think we should keep the defensive blocks and make LCOV ignore them. It would be nice if you leave some brief rationale for future developers.

sndth:
> @sndth Sorry for my late response. After some investigation and reflection, I think we should keep the defensive blocks and make LCOV ignore them. It would be nice if you leave some brief rationale for future developers.

Alright! I kept the defensive blocks, excluded them from LCOV, and added a brief rationale for future developers. Thanks for taking a look!😺

fktn-k (approved):
@sndth 
Looks good to me. Thanks a lot!

PR #535, opened 2026-07-27
Migrate to doctest

fktn-k:
This PR migrates the unit test framework from [Catch2](https://github.com/catchorg/Catch2) to [doctest](https://github.com/doctest/doctest) (v2.5.3).  
The main reason is that doctest still supports C++11 in the latest version while Catch2 has stopped supporting it since v3.

---

## Pull Request Checklist

## Please don't


tests/unit_test/test_custom_from_node.cpp
fktn-k, line 48: A note for myself: Add `operator <<` for this struct as the other ones in this file. Doctest requires it to make a log especially when an old GCC compiler is used.

tests/unit_test/CMakeLists.txt
```diff
@@ -127,10 +140,13 @@ target_compile_options(
       -Wall -Wextra -Werror -pedantic -Wdeprecated
       -Wno-c++98-compat -Wno-c++98-compat-pedantic
       -Wno-deprecated-declarations # for testing deprecated functions
+      -Wno-unused-comparison # for testing throwable operators
+      -Wno-unused-variable # for testing against old compilers
```
fktn-k, line 144: Check if doctest provides any preprocessor macro to suppress it since it seems to be caused by some part of the doctest implementation.

Issue #536, opened 2026-08-07
Memory-safety bugs in deserializer (stack/heap buffer over-read, invalid deque access) — reproduce on v0.4.3


AlexandrKhromov2005:
# Memory-safety bugs in fkYAML deserializer (reachable from untrusted YAML)

Fuzzing `fkyaml::node::deserialize(std::string)` with libFuzzer + ASan/UBSan surfaced a cluster of
memory-safety defects. All reproduce on the **latest release v0.4.3** (and v0.4.2). They are
reachable from any application that deserializes untrusted YAML.

Environment: clang 20, `-fsanitize=address,undefined,fuzzer`, x86-64 Linux. Harness just calls
`fkyaml::node::deserialize(input)` on the fuzzer bytes.

## 1. stack-buffer-overflow (READ) in `iterator_input_adapter::get_buffer_view_utf16`

**3-byte input** `01 00 0a` (hex). ASan:

```
ERROR: AddressSanitizer: stack-buffer-overflow ... READ of size 1
  #0 ... get_buffer_view_utf16() node.hpp:9407 (v0.4.3)   [9011 in v0.4.2]
  #1 ... get_buffer_view()       node.hpp:9301
  #2 ... basic_deserializer::deserialize(...)
  [32,64) 'input' <== Memory access at offset 64 overflows this variable
```

The leading `0x01 0x00` makes the input adapter treat the stream as UTF-16; for an odd byte count
the last 2-byte code unit is read one byte past the end of the buffer.

## 2. heap-buffer-overflow (READ) in `iterator_input_adapter::get_buffer_view_utf8`

Minimizes to **3 bytes**. ASan:

```
ERROR: AddressSanitizer: heap-buffer-overflow ... READ of size 1
  #0 ... get_buffer_view_utf8() node.hpp:8942 (v0.4.2)
  0x... is located 0 bytes after a 161-byte region
```

One-byte over-read at the end of the UTF-8 buffer view.

## 3. Invalid `std::deque::back()` on empty `m_context_stack` → SEGV

In `deserialize_scalar`, the explicit-key-separator branch dereferences
`m_context_stack.back().state` without checking that the stack is non-empty (node.hpp:8271 in
v0.4.2). A crafted document reaches it with an empty stack → invalid iterator deref → SEGV.
Note the very next scalar branch (a few lines below) *does* guard the same access with
`if (!m_context_stack.empty())`, so this looks like a missing guard:

```cpp
// current:
[...]

PR #543, opened 2026-08-09
Fix out-of-bounds read on an explicit key without a parent context

AlexandrKhromov2005:
This fixes an out-of-bounds read found while fuzzing `develop` after #540 (reported in #536).

`deserialize_node()` reads `m_context_stack.back()` when handling an `EXPLICIT_KEY_PREFIX` token without first checking that the stack is non-empty. An input such as `a  ? a` — a `? ` indicator following a completed scalar root — leaves the context stack empty at that point, so `back()` dereferences an empty `std::deque` and the process crashes with a SEGV (near-null read) instead of reporting a parse error. It reproduces in both Debug and `-DNDEBUG -O2` builds.

The fix guards the access and throws a `parse_error` for the misplaced key, mirroring the empty-stack handling already added at the mapping value separator in #540. A regression test with the minimal input is added to `test_fuzz_regression.cpp`.

I confirmed the input crashes before the change and returns a `parse_error` after it, and that the full unit test suite passes under `-fsanitize=address,undefined`.

---

## Pull Request Checklist

### DIFF
--- a/include/fkYAML/detail/input/deserializer.hpp
+++ b/include/fkYAML/detail/input/deserializer.hpp
@@ -408,6 +408,9 @@ private:
         do {
             switch (token.type) {
             case lexical_token_t::EXPLICIT_KEY_PREFIX: {
+                if FK_YAML_UNLIKELY (m_context_stack.empty()) {
+                    throw parse_error("An explicit key is not allowed in this context.", line, indent);
+                }
                 const bool needs_to_move_back = indent == 0 || indent < m_context_stack.back().indent;
                 if (needs_to_move_back) {
                     pop_to_parent_node(line, indent, [indent](const parse_context& c) {
--- a/include/fkYAML/detail/input/deserializer.hpp
+++ b/include/fkYAML/detail/input/deserializer.hpp
@@ -1294,9 +1294,11 @@ private:
     /// @param indent The indentation level of the target parent block mapping.
     template <typename Pred>
     void pop_to_parent_node(uint32_t line, uint32_t indent, Pred&& pred) {
+        // LCOV_EXCL_START
         if FK_YAML_UNLIKELY (m_context_stack.empty()) {
             throw parse_error("No parent block mapping is found.", line, indent);
         }
+        // LCOV_EXCL_STOP
 
         // LCOV_EXCL_START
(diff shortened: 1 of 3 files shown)


fktn-k (approved):
LGTM

include/fkYAML/detail/input/deserializer.hpp
```diff
@@ -408,6 +408,9 @@ class basic_deserializer {
         do {
             switch (token.type) {
             case lexical_token_t::EXPLICIT_KEY_PREFIX: {
+                if FK_YAML_UNLIKELY (m_context_stack.empty()) {
+                    throw parse_error("An explicit key is not allowed in this context.", line, indent);
+                }
```
fktn-k, line 413: With this change, `if FK_YAML_UNLIKELY (m_context_stack.empty())` (line 1298) never returns true in `pop_to_parent_node`.
After investigation, it's turned out that m_context_stack is always guranteed to be non-empty right before all the calls to this function.
So, could you surround the `if` check with `// LCOV_EXCL_START` and `// LCOV_EXCL_STOP` as follows so that the if check remains as-is for future development but is ignored for test coverage?

```cpp
template <typename Pred>
void pop_to_parent_node(uint32_t line, uint32_t indent, Pred&& pred) {
    // LCOV_EXCL_START
    if FK_YAML_UNLIKELY (m_context_stack.empty()) {
        throw parse_error("No parent block mapping is found.", line, indent);
    }
    // LCOV_EXCL_STOP

    // ...
}
```

fktn-k (changes requested):
@AlexandrKhromov2005 
Sorry for my change request after approvement.
I realized the coverage loss and found an existing check is unnecessary anymore.
Let me ask you for a small change for that.

include/fkYAML/detail/input/deserializer.hpp
AlexandrKhromov2005: Done — wrapped the check in `// LCOV_EXCL_START` / `// LCOV_EXCL_STOP` and kept it in place. You're right that it's unreachable now: with the new guard, every caller of `pop_to_parent_node` has a non-empty stack, so the check is dead but worth keeping as a guard for future changes. Amalgamation regenerated and in sync (`7af63c1`). No worries about the re-request after the approval.

fktn-k (approved):
LGTM. Thanks for the changes!

PR #544, opened 2026-08-10
Guard empty context stack in the key-separator and add_new_key paths

AlexandrKhromov2005:
Fuzzing `develop` after #540 surfaced two more crashes of the same class you fixed in #533/#538/#540 — an unguarded `m_context_stack.back()` when the context stack is empty (reported in #536).

- **`KEY_SEPARATOR` handler** — a `:` with no parent context (e.g. after a completed scalar root) reaches `m_context_stack.back().state` on an empty deque. Minimal input `0D 3B 00 0A 0A 4A 0A 0A 26 7C 0A 3A`.
- **`add_new_key()`** — a mapping key with no parent block-mapping context reaches `m_context_stack.back().indent` on an empty deque. Minimal input `26 78 00 61 5B 5D 0A 70 72 6F 7B 64 2B 3A 20 3E 72 6F 7B 64 2B 3A 7B 32 6A`.

Both dereference `back()` on an empty `std::deque` (UB; near-null read → SEGV in `-DNDEBUG` builds). Each is guarded with an `empty()` check that raises a `parse_error` for the misplaced token, mirroring the empty-stack handling added in #540. Two regression tests with the minimal inputs are added to `test_fuzz_regression.cpp`.

I confirmed both inputs crash before the change and return a `parse_error` after it (Debug and `-DNDEBUG -O2`), and that the full unit test suite passes under `-fsanitize=address,undefined`.

---

## Pull Request Checklist

### DIFF
--- a/include/fkYAML/detail/input/deserializer.hpp
+++ b/include/fkYAML/detail/input/deserializer.hpp
@@ -475,6 +475,9 @@ private:
                 continue;
             }
             case lexical_token_t::KEY_SEPARATOR: {
+                if FK_YAML_UNLIKELY (m_context_stack.empty()) {
+                    throw parse_error("A key separator is not allowed in this context.", line, indent);
+                }
                 if FK_YAML_UNLIKELY (m_context_stack.back().state == context_state_t::BLOCK_SEQUENCE_ENTRY) {
                     // empty mapping keys are not supported.
                     // ```yaml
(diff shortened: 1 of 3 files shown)


fktn-k (changes requested):
Thanks a lot for the contribution!
The changes you made seem fine, but could you please resolve the conflicts in test_fuzz_regression.cpp?

AlexandrKhromov2005:
Rebased onto `develop` and resolved the conflict in `test_fuzz_regression.cpp` (kept the #543 subcase and appended the two new ones). No code change to the guards. Full suite is green under ASan/UBSan.

fktn-k (approved):
LGTM. Thanks!

PR #545, opened 2026-08-10
Fix out-of-bounds read from a truncated BOM in the char encoding detector

AlexandrKhromov2005:
Fuzzing `develop` (reported in #536) found an out-of-bounds read reachable from ordinary `char`/`std::string` input.

`utf_encode_detector<char>::detect()` probes up to 4 leading bytes into an array padded with `0xFF`, then infers the encoding and a BOM. When the input is shorter than the BOM, the unread `0xFF` padding can spuriously complete a BOM pattern:

- a lone `0xFE` matches the UTF-16BE BOM (`FE FF`), and
- `00 00 FE` matches the UTF-32BE BOM (`00 00 FE FF`).

`detect()` then advanced `begin` by the full BOM length (2 or 4), moving it **past `end`**. The following read in `get_buffer_view_utf16()` / `get_buffer_view_utf32()` was an out-of-bounds read on the input buffer:

```
AddressSanitizer: heap-buffer-overflow ... READ of size 1
  in iterator_input_adapter<char*>::get_buffer_view_utf16() input_adapter.hpp:197
```

The fix tracks how many bytes were actually read and only skips a BOM that is fully present. Short inputs fall through to normal decoding, which reports a clean `invalid_encoding` instead of reading past the end — so valid BOMs behave exactly as before. A regression test with the 1-byte `0xFE` input is added.

I confirmed the input reads out of bounds before the change (heap-backed buffer, ASan) and throws `invalid_encoding` after it, in Debug and `-DNDEBUG -O2`, and that the full unit test suite (including the BOM/encoding tests) passes under `-fsanitize=address,undefined`.

The sibling `char8_t`/`char16_t`/`char32_t` detectors don't have this exposure: `char8_t` throws before advancing on any non-UTF-8 result, and the `char16_t`/`char32_t` BOM checks only use bytes from real code units.

---

## Pull Request Checklist

### DIFF
--- a/include/fkYAML/detail/encodings/utf_encode_detector.hpp
+++ b/include/fkYAML/detail/encodings/utf_encode_detector.hpp
@@ -104,27 +104,40 @@ struct utf_encode_detector<ItrType, enable_if_t<is_iterator_of<ItrType, char>::v
         std::array<uint8_t, 4> bytes {{}};
         bytes.fill(0xFFu);
         auto current = begin;
+        int num_read = 0;
         for (int i = 0; i < 4 && current != end; i++, ++current) {
             bytes[i] = static_cast<uint8_t>(*current); // NOLINT(cppcoreguidelines-pro-bounds-constant-array-index)
+            ++num_read;
         }
 
         bool has_bom = false;
         const utf_encode_t encode_type = detect_encoding_type(bytes, has_bom);
 
         if (has_bom) {
-            // skip reading the BOM.
+            // Skip reading the BOM, but only when it is fully present in the input. The probe above pads
+            // unread positions with 0xFF, so an input shorter than the BOM (e.g. a lone 0xFE) can spuriously
+            // match a BOM pattern; advancing past the bytes actually read would move `begin` past `end`.
+            int bom_size = 0;
             switch (encode_type) {
             case utf_encode_t::UTF_8:
-                std::advance(begin, 3);
+                bom_size = 3;
                 break;
             case utf_encode_t::UTF_16BE:
             case utf_encode_t::UTF_16LE:
-                std::advance(begin, 2);
+                bom_size = 2;
                 break;
             case utf_encode_t::UTF_32BE:
             case utf_encode_t::UTF_32LE:
-                std::advance(begin, 4);
+                bom_size = 4;
                 break;
+            }
+            if (num_read >= bom_size) {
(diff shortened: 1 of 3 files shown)


include/fkYAML/detail/encodings/utf_encode_detector.hpp
```diff
@@ -104,28 +104,36 @@ struct utf_encode_detector<ItrType, enable_if_t<is_iterator_of<ItrType, char>::v
...
                 break;
             case utf_encode_t::UTF_32BE:
             case utf_encode_t::UTF_32LE:
-                std::advance(begin, 4);
+                bom_size = 4;
                 break;
             }
+            if (num_read >= bom_size) {
+                std::advance(begin, bom_size);
+            }
```
fktn-k, line 136: To avoid false positives like the one mentioned in the comment above, it is better to return `utf_encode_t::UTF_8` rather than the detected `encode_type`.
```suggestion
            if (num_read >= bom_size) {
                std::advance(begin, bom_size);
            }
            else {
                return utf_encode_t::UTF_8;
            }
```

fktn-k (changes requested):
Thank you for the contribution.
I've reviewed your changes and requested some change for better consistency.
Also, please resolve the conflicts in test_fuzz_regression.cpp.

include/fkYAML/detail/encodings/utf_encode_detector.hpp
AlexandrKhromov2005: Done — returns `utf_encode_t::UTF_8` when the BOM is not fully present, so the spurious encoding is no longer propagated. Also rebased onto `develop` and resolved the `test_fuzz_regression.cpp` conflict. Thanks for the review.

fktn-k:
@AlexandrKhromov2005 
Could you resolve the conflict again..?  
Since the changes now look good to me, I will approve them after the conflict gets resolved.

AlexandrKhromov2005:
Rebased onto `develop` again (post-#544) and resolved the `test_fuzz_regression.cpp` conflict — kept the two subcases from #544 and appended the `truncated UTF-16 BOM` case. Full suite is green under ASan/UBSan; amalgamation re-synced. Thanks!

fktn-k (approved):
LGTM. Thanks a lot!

PR #547, opened 2026-08-10
Fix four out-of-bounds reads in the lexer and tag URI validation

### LINKED ISSUE #546: A few more crashes from fuzzing the lexer
### Description

Hello! I spent some time running the existing `tests/fuzz_test` harness under libFuzzer +
AddressSanitizer and came away with four crashes that I don't think anyone has reported yet
(maybe apart from `validate`?).

First, a quick note on the open PRs: I started by confirming that the bugs #544 and #545 fix are
real - the fuzzer finds both within a couple of seconds on a clean `develop`. To get any further
I applied those two patches locally, and everything below is what the fuzzer found after that.

So none of these four overlap with #544 or #545. They feel like the same family as #536.

## Four inputs
Every one of these is a tiny input that reads past the end of the buffer:

| Input | What happens?? |
| --- | --- |
| `!<` | Reads past the end while scanning a verbatim tag |
| `!!aaaaaaaaaaaaaaaaaaaaaaaaaaaaaa%41` | Reads past the end while validating a tag URI |
| `%TAG !  ` (two trailing spaces) | Reads past the end after the tag handle |
| `"\\\"` | Reads past the end, and keeps going - this one walks the heap |

## So, what's going on?
**`!<`**: the verbatim-tag branch advances the iterator to the end of the buffer, and then the
loop that follows advances it again *before* checking for the end. So it steps over the end
marker and never notices.

**`%41` in a tag URI**: `validate_octets()` advances the caller's iterator past the escape it
consumed, and then the outer loop advances it again. Besides running off the end, this means the
character right after a `%XX` escape is never validated. On `develop` today, `!!a%41^` is
accepted even though `^` isn't a legal URI character, while a perfectly valid `!!%41` gets
rejected - the byte being inspected is whatever happens to follow the buffer.

**`%TAG !  `**: after reading the tag handle the code skips whitespace, and if that whitespace
was the last thing in the buffer there's no check before it starts reading the tag prefix.

**`"\\\"`**: this is my favourite of the four. When the lexer sees a `"` it counts the
backslashes in front of it to decide whether the quote is escaped. That counting loop decrements
its pointer inside the loop condition, so it never looks at the very first character of the
scalar. An odd run of backslashes therefore gets counted as even, the lexer thinks it found the
closing quote, and hands back a token that's one byte short. Downstream, the scalar parser
unescapes past the end of that token and calls `str_view::remove_prefix()` with a length bigger
than the view - which does an unchecked `m_len -= n`, wraps the length around to something
enormous, and the next search happily scans until it hits unmapped memory.

(The comment right above that counting loop uses `"\\\""` as its example of the odd case, which
made me smile. 😆)

## Two things I noticed along the way
Not bugs exactly, more like observations, and I'm happy to be told these are intentional:

* **#544 adds a check at 2 of the ~40 places** where `deserializer.hpp` calls
  `m_context_stack.back()` without checking that the stack is non-empty. When I guarded all of
  them at once locally, the fuzzer stopped finding new ways into that same crash and moved on to
  other things. Might be worth a single accessor that throws, rather than a check per site.
* **`str_view::remove_prefix()` subtracts without checking.** `std::string_view` has the same
  precondition, so this isn't wrong as such - but it's what turned the fourth bug from "parses
  the wrong thing" into "scans the heap". A `FK_YAML_ASSERT(n <= m_len)` would have made it
  obvious in a debug build.

### Reproduction steps

```sh
printf '!<'                                   > 01.bin
printf '!!aaaaaaaaaaaaaaaaaaaaaaaaaaaaaa%41'  > 02.bin
printf '%TAG !  '                             > 03.bin
printf '"\\\\\\"'                             > 04.bin
```

Then run any of them through a build of `tests/fuzz_test/fuzzer_parse_yaml.cpp` with ASan.

One small snag if you want to reproduce on Windows: `tests/fuzz_test/CMakeLists.txt` uses the
clang spelling of the sanitizer flags, so the target won't configure with MSVC - even though
MSVC does ship both ASan and libFuzzer these days. I built the harness by hand with
`/fsanitize=address /fsanitize=fuzzer`. An `if(MSVC)` branch would make the target work there
too; let me know if that's worth a separate issue.

Expected result is to catch exceptions such as `parse_error`. Currently, the input accepts some invalid data

### Minimal code example

```Shell

```

### Error messages

```Shell

```

MSVC (VS2026); Windows 11 x64

### Library version

v0.4.3

### Validation

sndth:
This PR fixes four out-of-bounds reads in the lexer and tag URI validation, all found by
running the existing `tests/fuzz_test` harness under `libFuzzer` + `AddressSanitizer`.

Each one reproduces on unmodified `develop` and is reachable from
`fkyaml::node::deserialize()` with a tiny input:

| Input | Where |
| --- | --- |
| `!<` | `extract_tag_name()` steps past `m_end_itr` before the loop's end test can fire |
| `!!a%41` | `uri_encoding::validate()` advances the iterator twice per percent escape |
| `%TAG !  ` | `scan_tag_directive()` dereferences after `skip_white_spaces()` drained the buffer |
| `"\\\"` | backslash counting misses the token's first character, so the scalar parser gets a short token and `remove_prefix()` underflows |

These are separate from #544 and #545, so no overlap with either patch. I found them after
applying both locally so the fuzzer could get past those crashes.

**Two results change for inputs that didn't crash before:** `!!a%41^` is now rejected, since
the character after a percent escape is finally validated, and `!!%41` is now accepted, since a
percent escape may legitimately end a tag URI.

Fixes #546.

---

## Pull Request Checklist

## Please don't

### DIFF
--- a/include/fkYAML/detail/input/lexical_analyzer.hpp
+++ b/include/fkYAML/detail/input/lexical_analyzer.hpp
@@ -554,6 +554,11 @@ private:
         // extract a tag prefix.
         //
 
+        // skip_white_spaces() above may have consumed the rest of the input buffer.
+        if FK_YAML_UNLIKELY (m_cur_itr == m_end_itr) {
+            emit_error("invalid TAG directive is found.");
+        }
+
         m_token_begin_itr = m_cur_itr;
         const char* p_tag_prefix_begin = m_cur_itr;
         switch (*m_cur_itr) {
@@ -684,7 +689,9 @@ private:
         case '<':
             // Verbatim tags (!<TAG>)
             is_verbatim = true;
-            ++m_cur_itr;
+            if FK_YAML_UNLIKELY (++m_cur_itr == m_end_itr) {
+                emit_error("verbatim tag (!<TAG>) must be ended with \'>\'.");
+            }
             break;
         default:
             // Either local tags (!suffix) or named handles (!tag!suffix)
@@ -816,12 +823,18 @@ private:
                 // * even number of backslashes -> double quotation mark IS NOT escaped (e.g., "\\"")
                 uint32_t backslash_counts = 0;
                 const char* p = m_token_begin_itr + (pos - 1);
-                do {
-                    if (*p-- != '\\') {
+                for (;;) {
+                    if (*p != '\\') {
                         break;
                     }
                     ++backslash_counts;
-                } while (p != m_token_begin_itr);
+                    if (p == m_token_begin_itr) {
+                        // the first character of the token has just been counted. Stopping here
+                        // also keeps `p` from being decremented past the beginning of the token.
+                        break;
+                    }
+                    --p;
+                }
                 is_closed = ((backslash_counts & 1u) == 0); // true: even, false: odd
             }
 
--- a/include/fkYAML/detail/encodings/uri_encoding.hpp
+++ b/include/fkYAML/detail/encodings/uri_encoding.hpp
@@ -37,6 +37,11 @@ public:
                     return false;
                 }
 
+                // validate_octets() advances `current` past the last octet it consumed. Without
+                // moving it back, the loop's own ++current skips the character which follows the
+                // escape sequence, and, when the escape ends the sequence, moves `current` one
+                // past `end` so that the loop condition never holds and reads out of bounds.
+                --current;
                 continue;
             }
 
--- a/single_include/fkYAML/node.hpp
+++ b/single_include/fkYAML/node.hpp
@@ -1310,6 +1310,11 @@ public:
                     return false;
                 }
 
+                // validate_octets() advances `current` past the last octet it consumed. Without
+                // moving it back, the loop's own ++current skips the character which follows the
+                // escape sequence, and, when the escape ends the sequence, moves `current` one
+                // past `end` so that the loop condition never holds and reads out of bounds.
+                --current;
                 continue;
             }
 
@@ -3782,6 +3787,11 @@ private:
         // extract a tag prefix.
         //
 
+        // skip_white_spaces() above may have consumed the rest of the input buffer.
+        if FK_YAML_UNLIKELY (m_cur_itr == m_end_itr) {
+            emit_error("invalid TAG directive is found.");
+        }
+
         m_token_begin_itr = m_cur_itr;
         const char* p_tag_prefix_begin = m_cur_itr;
         switch (*m_cur_itr) {
@@ -3912,7 +3922,9 @@ private:
         case '<':
             // Verbatim tags (!<TAG>)
             is_verbatim = true;
-            ++m_cur_itr;
+            if FK_YAML_UNLIKELY (++m_cur_itr == m_end_itr) {
+                emit_error("verbatim tag (!<TAG>) must be ended with \'>\'.");
+            }
             break;
         default:
             // Either local tags (!suffix) or named handles (!tag!suffix)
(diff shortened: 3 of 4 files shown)


include/fkYAML/detail/input/lexical_analyzer.hpp
```diff
@@ -684,7 +689,9 @@ class lexical_analyzer {
         case '<':
             // Verbatim tags (!<TAG>)
             is_verbatim = true;
-            ++m_cur_itr;
+            if FK_YAML_UNLIKELY (++m_cur_itr == m_end_itr) {
+                emit_error("verbatim tag (!<TAG>) must be ended with \'>\'.");
+            }
```
fktn-k, line 694: Although I can't remember why `m_cur_itr` is incremented here, you can just remove `++m_cur_itr` to handle the error case when only `!<` is passed.

1. `if (++m_cur_itr == m_end_itr)` ends the do-while loops.
2. `if FK_YAML_UNLIKELY (last != '>')` at the beginning of `if (is_verbatim)` branch returns `true` and then the lexer would emit a parse error which says "verbatim tag (!<TAG>) must be ended with \'>\'.". Note that `tag_name.back()` can be safely called since it always has at least 2 characters "!<".

fktn-k (changes requested):
Thanks a lot for your contribution!
I've reviewed the changes and requested a minor change.

include/fkYAML/detail/input/lexical_analyzer.hpp
sndth: > Although I can't remember why `m_cur_itr` is incremented here, you can just remove
> `++m_cur_itr` to handle the error case when only `!<` is passed.

I tried that first. It does fix the crash exactly as you describe, but it breaks two existing
assertions:

```
test_lexical_analyzer_class.cpp(1611): FATAL ERROR:
REQUIRE_NOTHROW( token = lexer.get_next_token() )
THREW exception: "parse_error: invalid tag prefix (!) is found. (at line 0, column 2)"

[doctest] test cases: 371 | 370 passed | 1 failed
[doctest] assertions: 9450 | 9448 passed | 2 failed
```

Both are in `LexicalAnalyzer_Tag` / "valid tag names", from `!<!foo> tag` and
`!<!foo%2A%7C> tag`.

That also answers your question about why the increment is there: it makes the first character
after `<` skip the `switch` inside the scanning loop, and that `switch` has a `case '!'` which
emits `"invalid tag prefix (!) is found."` whenever `allows_another_tag_prefix` is false, which
it always is for a verbatim tag. So the increment is precisely what lets `!<!foo>` lex today.

> `if FK_YAML_UNLIKELY (last != '>')` at the beginning of `if (is_verbatim)` branch returns true
> […] Note that `tag_name.back()` can be safely called since it always has at least 2 characters

Both correct, I checked, with the increment removed, `!<` reaches the `last != '>'` check and
produces `verbatim tag (!<TAG>) must be ended with '>'.` with no ASan finding. The catch is that
`back()` is equally safe as the patch stands, since `m_cur_itr` is also at least 2 there, so that
property doesn't separate the two options; the assertions above do.

I also tried the obvious follow-up, remove the increment *and* skip the tag-prefix check inside
a verbatim tag (`!is_verbatim && !allows_another_tag_prefix`), which breaks the suite in the
opposite direction:

| Variant | Unit tests |
| --- | --- |
| This PR (`++m_cur_itr` + end check) | 371/371, 9460 assertions |
| Remove `++m_cur_itr` | 370/371, 2 failures in "valid tag names" |
| Remove it, allow `!` when `is_verbatim` | 370/371, `!<!f!oo> tag` no longer throws |

So the suite currently pins "exactly one `!`, and only as the first character of the URI", and
that behaviour exists only because the increment skips that one character. Any version without it
changes which tags are valid, which is why I went with the explicit end check, the smallest
change that fixes the read without touching behaviour.

That pinned behaviour looks accidental rather than intended, though, and it predates this PR:
`uri_encoding::validate_character()` already accepts `!` as a valid sub-delim under RFC 3986, so
arguably the tag-prefix check shouldn't apply inside `!<...>` at all and verbatim tag content
should just be "a valid URI". That would mean updating the `!<!f!oo> tag` expectation at line
1636, so I'd rather not fold it into this PR, happy to open a separate issue, or to reshape the
guard here if you'd prefer. Just let me know!!

fktn-k: @sndth
Thanks for the detailed information, which convinced me that the changes in this PR are reasonable. I'll resolve this thread.
If you have an idea for better implementation for parsing a verbatim tag, please open a PR for it; otherwise I'm fine with leaving it as-is.

fktn-k (approved):
The changes look good to me.
Please rebase the branch onto the latest develop HEAD before merging it.

sndth:
It's done 😄

PR #554, opened 2026-08-15
Complete explicit mapping keys which are not followed by a value on the same line

### LINKED ISSUE #553: Explicit key mapping entries are dropped unless a value follows on the same line
### Description

A block mapping entry written with an explicit key (`? key`) is only completed when a `: value`
follows on the same line. Every other way of writing the entry, i.e. an omitted value, no `:` at
all, or a value that starts on a following line, drops the entry from the result without an error.
The same entries written with an implicit key (`key:`) parse correctly, so only the explicit-key
form is affected.

Per the YAML spec an omitted value is a null value, and the explicit-key form is not special in
that regard (https://yaml.org/spec/1.2.2/#822-block-mappings). PyYAML parses all of the documents
below without complaint.

Two of these inputs trip an assertion in a debug build before producing the wrong result:
`node.hpp: FK_YAML_ASSERT(m_context_stack.back().state == context_state_t::MAPPING_VALUE)`.
The rest produce the wrong result silently, in both debug and NDEBUG builds.

Found while fuzzing for the empty-context-stack crashes in #536, but this one is not a
memory-safety problem (it's silent data loss on valid input, so I'm reporting it separately).

### Reproduction steps

Deserialize a block mapping whose entry uses an explicit key and is not finished by a `: value` on
the same line, for example an omitted value:

```yaml
? a
:
```

The entry disappears. It also disappears when other entries follow it, when the mapping is nested,
when the `:` is left out entirely (`? a`), and when the value starts on the next line (`? a` / `:` /
`- 1`), so a document can lose one entry and keep the rest.

| input | expected | actual |
|---|---|---|
| `"? a\n:\n"` | `{a: null}` | `{}` |
| `"? a\n: \n"` | `{a: null}` | `{}` |
| `"? a\n:\nb: c\n"` | `{a: null, b: c}` | `{b: c}` |
| `"foo:\n  ? a\n  :\n"` | `{foo: {a: null}}` | `{foo: {}}` |
| `"? a\n"` | `{a: null}` | `{}` |
| `"? a"` | `{a: null}` | `{}` |
| `"? a\n? b\n"` | `{a: null, b: null}` | `{}` |
| `"? a\n:\n- 1\n"` | `{a: [1]}` | `{}` |
| `"? a\n:\n  ? b\n  :\n"` | `{a: {b: null}}` | `{}` |
| `"? a\n: b\n"` | `{a: b}` | `{a: b}` (correct) |

The expected column is what PyYAML 6.0.3 produces for the same inputs. Of these, `"? a\n:\n"` and
`"? a\n:\n  ? b\n  :\n"` trip the debug assertion; the others are silent in both debug and NDEBUG
builds.

Additionally, `"? :\n"` (an empty explicit key with an omitted value) returns `{}` where PyYAML
raises a parser error. That one is the input that first tripped the assertion for me; whether it
should be an error or `{null: null}` is your call (the entries with a real key above are the
clear-cut part). Let me know which behavior you want and I'll match it.

### Minimal code example

```cpp
#include <iostream>
#include <fkYAML/node.hpp>

int main() {
  // explicit key, omitted value
  std::cout << fkyaml::node::serialize(fkyaml::node::deserialize(std::string("? a\n:\n")));
  // -> "{}"   expected "a: null"

  // implicit key, omitted value (parsed correctly)
  std::cout << fkyaml::node::serialize(fkyaml::node::deserialize(std::string("a:\n")));
  // -> "a: null"
}
```

### Error messages

```Shell

```

MSVC; Windows 11 x64

### Library version

v0.4.3

### Validation

sndth:
Fixes #553.

`basic_deserializer` keeps an explicit key (`? key`) in its own parse context until its value is
found, and only a `: value` on the same line ever completed the entry. Every other way of writing
the entry discarded the pending key node without an error, so the entry silently disappeared from
the result:

```yaml
? foo
? bar
baz: 123
# -> {baz: 123}, expected {foo: null, bar: null, baz: 123}

? qux
:
  corge
# -> {}, expected {qux: corge}
```

Two of these inputs also tripped `FK_YAML_ASSERT(m_context_stack.back().state == context_state_t::MAPPING_VALUE)`
in a debug build, since that branch assumed an omitted value can only happen for an implicit key.
With the entries completed properly they no longer reach that assertion.

## The change

There are two ways a pending explicit key entry gets completed, so this adds a helper for each and
calls them where the key used to be discarded:

- `add_explicit_key_with_null_value()` for a key with no value at all: at the end of a document,
  before another explicit key at the same indentation, before a mapping entry with an implicit key
  at the same indentation, and in the omitted-value branch of the key separator handling.
- `add_explicit_key_with_empty_value()` for a key whose value begins on a following line, which
  then parses into the value node as usual. The same-line path was already doing this inline, so it
  now shares the helper. Rebased onto the merged #552, it reads the parent context through
  `current_context()` so the empty-stack guard added there still applies after the key context is
  popped.

The third commit replaces the assertion in that same branch with a `parse_error`. It asserted that
the context reaching the omitted-value path is a mapping value context, which malformed input like
`? foo` / `:` / `:` violates: it aborts a debug build and, under `NDEBUG`, parses on from the
unexpected state. The only other contexts that reach it are `BLOCK_MAPPING` and
`BLOCK_MAPPING_EXPLICIT_VALUE`, and every document doing so in the corpus below is one PyYAML
rejects as malformed, so no valid document changes behavior.

Note that this error is consistent with how empty keys are handled today (`: a` / `: b` is already
rejected). If empty keys become supported the way you describe in #553, that input becomes valid
(`{foo: null, null: null}`) and this error will need revisiting. I would rather flag it now than
have it look like an oversight later.

## Verification

- The unit test suite passes: 371 test cases, 9525 assertions, with 4 new subcases in
  `test_deserializer_class.cpp` covering keys with omitted values, keys with no key separator at
  all, keys whose values begin on the following lines (scalar, block sequence at both
  indentations, nested explicit key), and the malformed key separator above.
- I also ran a differential test against PyYAML 6.0.3 over 16093 generated documents built from
  explicit-key line patterns, compared structurally: fkYAML's output is loaded back with PyYAML and
  checked against PyYAML's parse of the same input. Against `develop`, **1023 documents that were
  parsed incorrectly are now correct**, and nothing regressed except the class below.

## One behavior change worth pointing out 🤔

30 documents in that corpus now raise `parse_error: Detected duplication in mapping keys` where
they used to parse "successfully". These are documents where the same key appears twice, e.g.

```yaml
? foo
:
foo: 123
```

The explicit entry used to be dropped, so the duplicate never became visible; now both entries
reach the mapping and the duplicate is reported. That matches how fkYAML already treats
`foo: 1` / `foo: 2`, so I believe it is the correct outcome, but it is a visible change for anyone
whose input contains such a duplicate.

## Not in this PR!

Following your answer in #553 about empty keys, `"? :\n"` should end up as `{{null: null}: null}`
and `":\n"` as `{null: null}`. This PR moves `"? :\n"` from `{}` to `{null: null}`, which is closer
but not that, and it leaves `":\n"` as `{"": null}`, an empty **string** key, which is what fkYAML
does today and what an existing unit test asserts.

Getting to the behavior you described means supporting empty keys across block and flow, which is
currently missing in more places than this one: `: a` / `: b`, `{ : }` and `{ ? foo :, : bar }`
(spec example 7.3) are all rejected, and `[ : empty key ]` drops the key. That is a separate change
which also flips documented behavior (`""` to `null`), so I would rather do it in its own PR
against its own issue than widen this one. I did not lock `"? :\n"` into a test here, so nothing in
this PR stands in the way of that.

---

## Pull Request Checklist

## Please don't

### DIFF
--- a/include/fkYAML/detail/input/deserializer.hpp
+++ b/include/fkYAML/detail/input/deserializer.hpp
@@ -587,19 +587,35 @@ private:
                     }
 
                     if (indent <= m_context_stack.back().indent) {
-                        FK_YAML_ASSERT(m_context_stack.back().state == context_state_t::MAPPING_VALUE);
+                        if (m_context_stack.back().state == context_state_t::BLOCK_MAPPING_EXPLICIT_KEY) {
+                            // An explicit key can omit its value as well, in which case the entry must still be
+                            // added to the parent mapping.
+                            // ```yaml
+                            // ? foo
+                            // :
+                            // bar: baz
+                            // # -> {foo: null, bar: baz}
+                            // ```
+                            basic_node_type key_node = std::move(*m_context_stack.back().p_node);
+                            m_context_stack.pop_back();
+                            m_context_stack.back().p_node->as_map().emplace(std::move(key_node), basic_node_type());
+                            mp_current_node = m_context_stack.back().p_node;
+                        }
+                        else {
+                            FK_YAML_ASSERT(m_context_stack.back().state == context_state_t::MAPPING_VALUE);
 
-                        // Mapping values can be omitted and are considered to be null.
-                        // ```yaml
-                        // foo:
-                        // bar:
-                        //   baz:
-                        // qux:
-                        // # -> {foo: null, bar: {baz: null}, qux: null}
-                        // ```
-                        pop_to_parent_node(line, indent, [indent](const parse_context& c) {
-                            return (c.state == context_state_t::BLOCK_MAPPING) && (indent == c.indent);
-                        });
+                            // Mapping values can be omitted and are considered to be null.
+                            // ```yaml
+                            // foo:
+                            // bar:
+                            //   baz:
+                            // qux:
+                            // # -> {foo: null, bar: {baz: null}, qux: null}
+                            // ```
+                            pop_to_parent_node(line, indent, [indent](const parse_context& c) {
+                                return (c.state == context_state_t::BLOCK_MAPPING) && (indent == c.indent);
+                            });
+                        }
                     }
 
                     // defer checking the existence of a key separator after the following scalar until the next
--- a/include/fkYAML/detail/input/deserializer.hpp
+++ b/include/fkYAML/detail/input/deserializer.hpp
@@ -624,7 +624,15 @@ private:
                         // # -> {foo: null, bar: baz}
                         // ```
                         if (!add_explicit_key_with_null_value()) {
-                            FK_YAML_ASSERT(m_context_stack.back().state == context_state_t::MAPPING_VALUE);
+                            if FK_YAML_UNLIKELY (m_context_stack.back().state != context_state_t::MAPPING_VALUE) {
+                                // The key separator does not follow a mapping key, for example:
+                                // ```yaml
+                                // ? foo
+                                // :
+                                // :
+                                // ```
+                                throw parse_error("A key separator is not allowed in this context.", line, indent);
+                            }
 
                             // Mapping values can be omitted and are considered to be null.
                             // ```yaml
--- a/include/fkYAML/detail/input/deserializer.hpp
+++ b/include/fkYAML/detail/input/deserializer.hpp
@@ -291,6 +291,13 @@ private:
             last_type == lexical_token_t::END_OF_BUFFER || last_type == lexical_token_t::END_OF_DIRECTIVES ||
             last_type == lexical_token_t::END_OF_DOCUMENT);
 
+        // An explicit key at the end of a document has no value either.
+        // ```yaml
+        // ? foo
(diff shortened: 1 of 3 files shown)


fktn-k (approved):
LGTM

fktn-k:
@sndth   
Thanks a lot for the contribution!
Regarding the "One behavior change worth pointing out 🤔" section in the description, I don't regard this fix as a breaking change.  
Since a mapping entry with both a explicit key and value was successfully deserialized into the parsed node tree, the previous behavior with the value omitted was just a bug.  
Also, library users would have no reason to depend on it rather than reporting it as a bug.

PR #558, opened 2026-08-16
Support empty keys in block and flow mappings

### LINKED ISSUE #557: Empty keys are only supported at the document root, and as an empty string instead of null
### Description

An empty key (`: value` with nothing before the `:`) is only handled at the very start of a
document, and it produces an entry whose key is an empty **string**. Everywhere else the key
separator is rejected, and inside a flow mapping it is ignored, which silently drops the entry.

Per the YAML specification an empty key is an empty node, i.e. null, and it is allowed anywhere a
mapping entry is (https://yaml.org/spec/1.2.2/#822-block-mappings). Spec example 8.18 uses one.

As discussed in #553, changing the key from an empty string to null is a breaking change for anyone
reading such an entry as `node[""]`, so it would need a v0.5.0 release. A quoted empty key
(`"": value`) is a genuine empty string and is not affected.

### Reproduction steps

Deserialize a mapping whose key is empty, in any position other than the first line of a document.

| input | expected | actual on `develop` (57997a8) |
|---|---|---|
| `": foo"` | `{null: foo}` | `{"": foo}` - key is an empty string |
| `":"` | `{null: null}` | `{"": null}` |
| `"key: value\n: empty key"` | `{key: value, null: 'empty key'}` | `parse_error: Detected invalid indentation.` |
| `":\nb: 1"` | `{null: null, b: 1}` | `parse_error: Detected invalid indentation.` |
| `"- : foo"` | `[{null: foo}]` | `parse_error: mapping key should not be empty.` |
| `"{key: value, : empty key}"` | `{key: value, null: 'empty key'}` | `parse_error: The ":" mapping value indicator must be followed after a mapping key.` |
| `"\"\": foo"` | `{"": foo}` | `{"": foo}` (correct, a quoted key stays a string) |

In the yaml-test-suite these are the cases `NKF9` (Empty keys in block and flow mapping), `PW8X`
(Anchors on Empty Scalars) and `S3PD` (Spec Example 8.18. Implicit Block Mapping Entries).

### Minimal code example

```cpp
#include <iostream>
#include <fkYAML/node.hpp>

int main() {
  // handled, but the key is an empty string rather than null
  std::cout << fkyaml::node::serialize(fkyaml::node::deserialize(std::string(": foo\n")));
  // -> "\"\": foo"   expected "null: foo"

  // rejected
  fkyaml::node::deserialize(std::string("key: value\n: empty key\n"));
}
```

### Error messages

```Shell

```

MSVC; Windows 11

### Library version

v0.4.3

### Validation

sndth:
This PR fixes #557.

An empty key was only handled at the very start of a document, and it produced an entry whose key
is an empty **string**. Everywhere else the key separator was rejected, and inside a flow mapping it
was ignored, which dropped the entry:

```yaml
key: value
: empty key                  # parse_error: Detected invalid indentation
- : foo                      # parse_error: mapping key should not be empty
{ key: value, : empty key }  # parse_error: The ":" mapping value indicator must be
                             # followed after a mapping key
```

## The change

`add_empty_key_entry()` adds an entry with a null key and moves to the token after the separator.
It is called wherever a key separator turns up with no key before it: at the document root, at the
start of a line in a block mapping, in a block sequence entry, and in a flow mapping. The
deliberate `mapping key should not be empty` rejection for block sequence entries is gone.

The value of an empty key can be omitted too, so the helper also moves back to the parent mapping
when the following token is not indented deeper - the same rule a normal key already follows:

```yaml
:
foo: bar
# -> {null: null, foo: bar}

:
  foo: bar
# -> {null: {foo: bar}}
```

| input | before | after |
|---|---|---|
| `": foo"` | `{"": foo}` | `{null: foo}` |
| `":"` | `{"": null}` | `{null: null}` |
| `"key: value\n: empty key"` | `parse_error` | `{key: value, null: 'empty key'}` |
| `":\nb: 1"` | `parse_error` | `{null: null, b: 1}` |
| `"- : foo"` | `parse_error` | `[{null: foo}]` |
| `"{key: value, : empty key}"` | `parse_error` | `{key: value, null: 'empty key'}` |
| `"\"\": foo"` | `{"": foo}` | unchanged, a quoted key stays a string |

## Breaking change

As you noted in #553, an empty key is now a null node rather than an empty string, so code reading
such an entry as `node[""]` has to read `node[nullptr]` instead. Two existing subcases in
`test_deserializer_class.cpp` asserted the old behavior and are updated here. I did not touch any
version macro, assuming the v0.5.0 bump happens at release time as usual.

---

## Pull Request Checklist

## Please don't

### DIFF
--- a/include/fkYAML/detail/input/deserializer.hpp
+++ b/include/fkYAML/detail/input/deserializer.hpp
@@ -253,10 +253,7 @@ private:
             apply_node_properties(root);
             m_context_stack.emplace_back(
                 lexer.get_lines_processed(), lexer.get_last_token_begin_pos(), context_state_t::BLOCK_MAPPING, &root);
-            add_new_key(basic_node_type(""), line, indent);
-            token = lexer.get_next_token();
-            line = lexer.get_lines_processed();
-            indent = lexer.get_last_token_begin_pos();
+            add_empty_key_entry(lexer, token, line, indent);
             break;
         case lexical_token_t::BLOCK_LITERAL_SCALAR:
         case lexical_token_t::BLOCK_FOLDED_SCALAR:
@@ -500,15 +497,32 @@ private:
                 if FK_YAML_UNLIKELY (m_context_stack.empty()) {
                     throw parse_error("A key separator is not allowed in this context.", line, indent);
                 }
-                if FK_YAML_UNLIKELY (m_context_stack.back().state == context_state_t::BLOCK_SEQUENCE_ENTRY) {
-                    // empty mapping keys are not supported.
+                if (m_context_stack.back().state == context_state_t::BLOCK_SEQUENCE_ENTRY) {
+                    // The entry is a mapping whose first key is empty.
                     // ```yaml
                     // - : foo
+                    // # -> [{null: foo}]
                     // ```
-                    throw parse_error("mapping key should not be empty.", line, indent);
+                    *mp_current_node = basic_node_type::mapping();
+                    apply_directive_set(*mp_current_node);
+                    m_context_stack.emplace_back(line, indent, context_state_t::BLOCK_MAPPING, mp_current_node);
+                    add_new_key(basic_node_type(), line, indent);
+
+                    token = lexer.get_next_token();
+                    indent = lexer.get_last_token_begin_pos();
+                    line = lexer.get_lines_processed();
+                    continue;
                 }
 
                 if (m_flow_context_depth > 0) {
+                    if (m_context_stack.back().state != context_state_t::MAPPING_VALUE) {
+                        // No key precedes this separator, so the entry has an empty key.
+                        // ```yaml
+                        // { : foo }
+                        // # -> {null: foo}
+                        // ```
+                        add_new_key(basic_node_type(), line, indent);
+                    }
                     break;
                 }
 
@@ -1281,6 +1295,18 @@ private:
                 }
 
                 if (m_context_stack.back().state != context_state_t::BLOCK_MAPPING_EXPLICIT_KEY) {
+                    const parse_context& cur_context = m_context_stack.back();
+                    if (cur_context.state == context_state_t::BLOCK_MAPPING && cur_context.indent == indent) {
+                        // A key separator which begins a line belongs to an entry with an empty key.
+                        // ```yaml
+                        // foo: bar
+                        // : baz
+                        // # -> {foo: bar, null: baz}
+                        // ```
+                        add_empty_key_entry(lexer, token, line, indent);
+                        return;
+                    }
+
                     pop_to_parent_node(line, indent, [indent](const parse_context& c) {
                         return c.state == context_state_t::BLOCK_MAPPING_EXPLICIT_KEY && indent == c.indent;
                     });
@@ -1392,6 +1418,35 @@ private:
         }
         // LCOV_EXCL_STOP
         return m_context_stack.back();
+    }
+
+    /// @brief Adds a mapping entry whose key is empty and moves to the token which follows it.
+    /// @note
+    /// An empty key is a null node. Its value can be omitted as well, in which case the following token
+    /// belongs to the parent mapping rather than to this entry.
+    /// ```yaml
+    /// :
+    /// foo: bar
+    /// # -> {null: null, foo: bar}
+    /// ```
+    /// @param lexer The lexical analyzer to be used.
+    /// @param token The storage for the token which follows the key separator.
+    /// @param line The line of the key separator. Updated to the line of the following token.
+    /// @param indent The indentation width of the key separator. Updated for the following token.
+    void add_empty_key_entry(lexer_type& lexer, lexical_token& token, uint32_t& line, uint32_t& indent) {
+        const uint32_t key_line = line;
+        const uint32_t key_indent = indent;
+        add_new_key(basic_node_type(), line, indent);
+
+        token = lexer.get_next_token();
+        line = lexer.get_lines_processed();
+        indent = lexer.get_last_token_begin_pos();
+
+        if (line > key_line && indent <= key_indent) {
+            pop_to_parent_node(line, indent, [key_indent](const parse_context& c) {
+                return c.state == context_state_t::BLOCK_MAPPING && key_indent == c.indent;
+            });
+        }
     }
 
     /// @brief Adds an entry for an explicit key and makes its value node the current node.
--- a/include/fkYAML/detail/input/deserializer.hpp
+++ b/include/fkYAML/detail/input/deserializer.hpp
@@ -296,12 +296,20 @@ private:
             last_type == lexical_token_t::END_OF_BUFFER || last_type == lexical_token_t::END_OF_DIRECTIVES ||
             last_type == lexical_token_t::END_OF_DOCUMENT);
 
-        // An explicit key at the end of a document has no value either.
+        // An explicit key at the end of a document has no value either. Its own contents may have left
+        // more contexts on the stack, so those are unwound first.
         // ```yaml
         // ? foo
-        // # -> {foo: null}
+        // ? bar: baz
+        // # -> {foo: null, {bar: baz}: null}
         // ```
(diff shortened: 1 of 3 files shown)


sndth:
I added the `? :` patch here as a second commit.

The reasoning follows yours: `? ` introduces an explicit key, and a `:` on the same line belongs to
the contents of that key rather than to the entry the key is part of. So that separator now turns
the key node into a mapping and adds an entry with an empty key to it, exactly the way a scalar
does in `? foo: bar`. Two follow-on adjustments were needed: the separator which later terminates
the explicit key has to find its context below the ones the key's own contents left on the stack,
and completing a pending explicit key at the end of a document has to unwind those contexts too.

| input | before | after |
|---|---|---|
| `? :` | `{null: null}` | `{{null: null}: null}` |
| `? : foo` | `{foo: null}` | `{{null: foo}: null}` |
| `? foo: bar` | `{}` | `{{foo: bar}: null}` |
| `? :` / `: baz` | `parse_error` | `{{null: null}: baz}` |

The last two rows come along for free, and I think they are worth pointing out: `? foo: bar` was
silently dropping the entire entry, and `? :` followed by `: baz` was rejected. Both reproduce on
`develop` without either commit of this PR, so they are pre-existing rather than something the
first commit introduced.

I'll open the two issues you suggested, for an omitted value in a flow mapping (`{foo: }`) and for
single pair mappings in a flow collection (`[foo: 1]`), and link them here.

include/fkYAML/detail/input/deserializer.hpp
```diff
@@ -491,15 +496,52 @@ class basic_deserializer {
...
+                    indent = lexer.get_last_token_begin_pos();
+                    line = lexer.get_lines_processed();
+                    continue;
+                }
+
+                {
+                    const parse_context& cur_context = m_context_stack.back();
+                    const bool is_explicit_key_content =
+                        cur_context.state == context_state_t::BLOCK_MAPPING_EXPLICIT_KEY && cur_context.line == line &&
+                        cur_context.indent < indent;
```
fktn-k: Is this check necessary? I can't imagine a lexical token with the same or greater indent given the lexer only reads the input forwardly.

include/fkYAML/detail/input/deserializer.hpp
```diff
@@ -491,15 +496,52 @@ class basic_deserializer {
...
+                    if (is_explicit_key_content) {
+                        // The content of an explicit key begins with a key separator, so the key is a
+                        // mapping whose first entry has an empty key.
+                        // ```yaml
+                        // ? :
+                        // # -> {{null: null}: null}
```
fktn-k: Strictly speaking, this comment seems misleading since it's unknown whether the value is also empty at this timing. (It will be known in `add_empty_key_entry` a few lines below.)  
So, it would be nice if the snippet reflects the leading comment which describes the context perfectly.

fktn-k (changes requested):

include/fkYAML/detail/input/deserializer.hpp
sndth: You're right! It's dead weight. The context stores the position of the `? ` token, so a key separator
on that same line can only come after it, which makes the indent comparison implied by
`cur_context.line == line`. I removed it.

include/fkYAML/detail/input/deserializer.hpp
sndth: Agreed, that snippet described the end result rather than what is known here. At this point only the
key is decided: its contents begin with a key separator, so it is a mapping whose first entry has an
empty key. Whether that entry has a value, and whether the explicit key itself has one, is settled
later. Reworded the comment and made the snippet point at the separator instead of showing the final
node.

include/fkYAML/detail/input/deserializer.hpp
```diff
@@ -1379,6 +1450,35 @@ class basic_deserializer {
...
+    void add_empty_key_entry(lexer_type& lexer, lexical_token& token, uint32_t& line, uint32_t& indent) {
+        const uint32_t key_line = line;
+        const uint32_t key_indent = indent;
+        add_new_key(basic_node_type(), line, indent);
+
+        token = lexer.get_next_token();
+        line = lexer.get_lines_processed();
+        indent = lexer.get_last_token_begin_pos();
+
+        if (line > key_line && indent <= key_indent) {
```
fktn-k, line 1490: I found this branch seems not covered in the unit tests.  
Is the `indent <= key_indent` check necessary in the first place?

fktn-k (changes requested):

include/fkYAML/detail/input/deserializer.hpp
sndth: Oh, so sorry! I completely forgot about the unit tests 🥲

> Is the `indent <= key_indent` check necessary in the first place?

The `indent <= key_indent` check is necessary though. It decides whether the token after the
separator is a sibling of the empty key or its value:

```yaml
:
bar: baz     # -> {null: null, bar: baz}, the empty key has no value

:
  bar: baz   # -> {null: {bar: baz}}, the mapping is the value of the empty key
```

Dropping the check makes the second form fail with `bad indentation of a mapping entry`, since the
context is popped before the nested mapping is attached. The two new subcases pin exactly that
difference.

fktn-k: Oh, I overlooked the cases. Thank you for the explanation!

fktn-k:
@sndth  
Although the changes look good to me, I will keep this PR unmerged until the next 0.4.4 gets released for these reasons:
- #563 (the release blocker) is about to be merged.
- We have many changes irrelevant to the breaking change for the next release.
- Empty keys are a relatively minor feature in terms of common YAML usages.

fktn-k (changes requested):
Could you update the branch onto the develop HEAD?  
Since #563 has been merged, all the CI tests should pass successfully now.

sndth:
> Could you update the branch onto the develop HEAD? Since #563 has been merged, all the CI tests should pass successfully now.

Of course, it's done!

fktn-k:
@sndth  
I investigated the newly failed case (`Y79Y_07`, `? -\n:\t-`).  
The previous parser throws a `parse_error` but on a wrong part of the YAML, the exception says:
```
fkYAML/tests/unit_test/test_deserializer_class.cpp:3465: FATAL ERROR: REQUIRE_NOTHROW( root = deserializer.deserialize(fkyaml::detail::input_adapter("? -\n:\t-")) ) THREW exception: "parse_error: mapping key should not be empty. (at line 1, column 0)"
```

The `? -` part is syntactically correct and is one of what you've fixed in this PR.  
Then the parser mistakenly admits the `:\t` part.  
The YAML specification clearly defines here a key separator as `a colon and a space`, not a white space which includes a tab.  
https://yaml.org/spec/1.2.2/#61-indentation-spaces

So, **the failure is completely irrelevant to the changes in this PR**.  
That is why I would like to merge this PR even with the failure.  
What would you think about it?

sndth:
> @sndth I investigated the newly failed case (`Y79Y_07`, `? -\n:\t-`). The previous parser throws a `parse_error` but on a wrong part of the YAML, the exception says:
> 
> ```
> fkYAML/tests/unit_test/test_deserializer_class.cpp:3465: FATAL ERROR: REQUIRE_NOTHROW( root = deserializer.deserialize(fkyaml::detail::input_adapter("? -\n:\t-")) ) THREW exception: "parse_error: mapping key should not be empty. (at line 1, column 0)"
> ```
> 
> The `? -` part is syntactically correct and is one of what you've fixed in this PR. Then the parser mistakenly admits the `:\t` part. The YAML specification clearly defines here a key separator as `a colon and a space`, not a white space which includes a tab. https://yaml.org/spec/1.2.2/#61-indentation-spaces
> 
> So, **the failure is completely irrelevant to the changes in this PR**. That is why I would like to merge this PR even with the failure. What would you think about it?

Yes, I'm fine with it.

fktn-k (approved):
LGTM

PR #562, opened 2026-08-19
Integrate with the YAML test suite

fktn-k:
This PR integrates tests with the YAML test suite, the official test suite from the YAML organization with a lot of valid/invalid test data.  
You can run the tests with the following commands:
```bash
$ cd /path/to/fkYAML
$ cmake -S . -B build -DCMAKE_BUILD_TYPE=Debug -DFK_YAML_USE_YAML_TEST_SUITE=ON
$ cmake --build build
$ ctest --test-dir build --output-on-failure -R yaml_test_suite_
```

Also, a Python script is added so we can easily enumerate a list of failed cases as a report in the Markdown format.  
You can use it by running the following commands:
```bash
# Assume you can run the YAML test suite tests with the `build` directory
$ cd /path/to/fkYAML
$ python3 ./scripts/report_yaml_test_suite_failures.py --test-dir build --output ./report.md
```

Since the test suite doesn't provide its test runner, I've implemented one.
It behaves like this:
1. CMake enumerate test case directories in the test suite and generates from them a list of `doctest` test cases for both YAML/JSON inputs into a generated header file.
2. CTest runs each generated `doctest` test case in the following steps:
  - If the case is an error case, the runner checks if fkYAML fails to parse the input.
  - If the case is a normal case, the runner checks if fkYAML succeeds to parse the input.
    1. The runner parses `test.event` for the test case to generate a set of necessary validators.
    2. The runner validates the deserialization output from fkYAML using the validators.

---

## Pull Request Checklist

## Please don't


sndth:
Read through the runner, the workflow and the report script. The design LGTM, and
validating against `test.event` is a level above what I had prototyped: it covers tags, anchors
(with the registry so aliases resolve to their anchor) and scalar styles, which a JSON comparison
cannot reach. Splitting each case into its own ctest test is also what makes the next point
survivable.

**The report drops cases which abort the process**: `parse_failures()` only matches `***Failed` and
`(Failed)`, but ctest also reports `***Exception: Child aborted` / `(Subprocess aborted)`,
`(SEGFAULT)`, `(Timeout)` and `(Not Run)`. The workflow builds in Debug, so `FK_YAML_ASSERT` is
live, and four suite cases abort today on `develop` through
`FK_YAML_ASSERT(tag_type != tag_t::SEQUENCE && tag_type != tag_t::MAPPING)` in `scalar_parser.hpp`:

| case | label |
|---|---|
| `6JWB` | Tags for Block Objects |
| `735Y` | Spec Example 8.20. Block Node Types |
| `BU8L` | Node Anchor and Tag on Seperate Lines |
| `H7J7` | Node anchor not indented (an error case) |

A `!!map` or `!!seq` tag on a block sequence entry, or a tag on a line of its own, ends up routed
into the scalar parser. So the most severe results are exactly the ones the report would omit today,
and a reader would conclude the library returns wrong contents rather than that it crashes on a spec
example. Widening the regexes to any non-passed status would cover it.

**One thought on the workflow**: With `continue-on-error: true` the job can never fail, so a change
which breaks ten more cases looks the same as one which fixes ten, unless someone diffs the two
artifacts by hand. A committed list of currently failing case IDs plus a step which fails only when
a new ID appears would give a ratchet without blocking on the existing backlog.

fktn-k:
@sndth  
Thank you for the feedbacks!  
> The report drops cases which abort the process: 

Oh.. I didn't noticed that. I'll fix it!

> One thought on the workflow

Your idea is great! I'll try to implement the mechanism.

fktn-k:
I will not include the changes in this PR for the next 0.4.4 release since there are some more to do for this feature.

fktn-k:
@sndth  
> **The report drops cases which abort the process**:

I widened the regex to non-passed statuses.  
As your said, that covered all the mistakenly dropped failure cases.  
Thanks!  

> **One thought on the workflow**

I implemented a mechanism which follows your suggestion: Only if any new ID appears in the failure lists, the workflow fails.  

Could you check the last 2 commits which address them?

sndth:
Both LGTM and I checked them rather than just read them 😸

**The regex widening**: I ran `parse_failures()` against synthetic ctest output covering every status
I could think of, and it catches all of them now: `***Failed`, `***Exception: Child aborted`,
`***Exception: SegFault`, `***Timeout`, `***Not Run`, while `Passed` is correctly skipped. So the four
assertion cases (`6JWB`, `735Y`, `BU8L`, `H7J7`) will show up in the report. I also checked the dot
padding, since the pattern needs at least one `.` between the name and the status: ctest aligns to the
longest test name in the run and even that one still gets dots, so there's no risk of a long generated
name dropping its line.

One small thing: `***Skipped` would now be counted as a failure. It can't happen today, since the
doctest CMake module doesn't set `SKIP_RETURN_CODE`, but with the new ratchet a skipped test would fail
the workflow, so it may be worth treating `Skipped` like `Passed`.

**The ratchet**: This is better than what I suggested. Rebuilding the baseline ref beats a list
committed in the repo, which would need updating and would rot. I like that the expensive path only
runs when the current run already failed, and the three edge cases are handled properly: no previous
commit, a baseline predating the integration (`Total Tests: 0`), and a non-PR non-push event. The
comparison semantics are right too, since `pull_request` builds the merge commit while the baseline is
`base.sha`, so the diff is the PR's own contribution rather than anything inherited from `develop`.

The one risk I'd flag is cost: on a failure the job builds and tests twice, and `FetchContent` pulls the
test suite into both build trees. With `-j 2` and 20 minutes that may get tight, and a shared
`FETCHCONTENT_BASE_DIR` or an `actions/cache` entry for the suite would take that pressure off?

And a general note on both: an unrecognised status line is dropped silently, which is asymmetric for a
gate (a new failure whose line doesn't match would pass the ratchet unnoticed). Counting lines which
match `Test #` but no status pattern, and failing loudly if that count isn't zero, would close it.

fktn-k:
@sndth  
Thank you for the review!  
I will merge this PR since we both agree on the basic design of this feature.  
While there are still a few areas that need improvement, I would rather think the benefits of using this feature for regression testing of future changes outweigh the risks you pointed out.  

So, I will create a separate PR to improve the regex and the ratchet.  
I would appreciate it if you could review those changes as well.

PR #563, opened 2026-08-19
Hold parse context node ownership in a unique_ptr

sndth:
Fixes mem leaks reported in #536.

A parse context either borrows a node from the result tree or owns one until that node is grafted
into the tree. That distinction was derived from the mutable `state` member, which the parsing code
rewrites while the node is still owned: the `default:` branch of the flow collection handlers turns
a `FLOW_MAPPING_KEY` context into a `FLOW_MAPPING` one, after which the destructor no longer freed
the node it still owned.

This holds the owned node in a `std::unique_ptr` member instead, makes parse contexts **move-only**
so ownership cannot be duplicated by a copy, and decides whether a closing flow collection is a
mapping key from that member rather than from `state`. The manual `delete` calls and the destructor
which freed a raw pointer based on the state are gone.

## Why not a flag 🤔

@AlexandrKhromov2005 prototyped the `owns_node` bool first and found it turns the leaks into
use-after-free. Running his branch under ASan shows why, and it is not a missing `owns_node = false`:

- `parse_context` stays copyable, so a copy duplicates the flag and with it the ownership of one
  pointer, and
- in the flow ending path ownership is handed to a context whose node was just swapped into the
  result tree, so the freed node is one the tree still points at.

A mutable bool can drift from reality exactly like the state inference did. With ownership in a
`unique_ptr` and the context move-only, a duplicated owner is a compile error rather than a runtime
UAF, which is the part a flag cannot give.

| | `owns_node` prototype | this PR |
|---|---|---|
| `{{{`, `? [`, `? []`, `{{{}},{` | crashes on the fourth | all clean! |
| 16093 generated documents under ASan | heap-use-after-free after 135 | 16093/16093, no reports! |

## This should also fix the red sanitizer job

`ci_test_clang_sanitizers` is currently failing on `develop`, and it is this bug: the regression
inputs added with #552, `{{{}},{` and `{{{}},[`, leak, so LeakSanitizer fails the test binary.
Measured with the CRT debug heap, all five of `{{{`, `? [`, `? []`, `{{{}},{` and `{{{}},[` leak on
`develop` and are clean with this change. I could not run Clang with LSan locally, so whether the
job goes green will be visible on this PR's own run rather than something I can promise.

---

## Pull Request Checklist

## Please don't

### DIFF
--- a/include/fkYAML/detail/input/deserializer.hpp
+++ b/include/fkYAML/detail/input/deserializer.hpp
@@ -11,6 +11,7 @@
 
 #include <algorithm>
 #include <deque>
+#include <memory>
 #include <vector>
 
 #include <fkYAML/detail/macros/define_macros.hpp>
@@ -64,11 +65,15 @@ class basic_deserializer {
     };
 
     /// @brief Context information set for parsing.
+    /// @note
+    /// A context either borrows a node which the result tree already owns, or owns a node of its own until that
+    /// node is grafted into the tree. Ownership is held by the owned_node member rather than derived from the
+    /// state member, which the parsing code rewrites while a node is still owned.
     struct parse_context {
         /// @brief Construct a new parse_context object.
         parse_context() = default;
 
-        /// @brief Construct a new parse_context object with non-default values for each parameter.
+        /// @brief Construct a new parse_context object which borrows a node owned by the result tree.
         /// @param line The current line. (count from zero)
         /// @param indent The indentation width in the current line. (count from zero)
         /// @param state The parse context type.
@@ -80,23 +85,26 @@ class basic_deserializer {
               p_node(p_node) {
         }
 
-        parse_context(const parse_context&) noexcept = default;
-        parse_context& operator=(const parse_context&) noexcept = default;
+        /// @brief Construct a new parse_context object which owns its node.
+        /// @param line The current line. (count from zero)
+        /// @param indent The indentation width in the current line. (count from zero)
+        /// @param state The parse context type.
+        /// @param node The node owned by this context.
+        parse_context(
+            uint32_t line, uint32_t indent, context_state_t state, std::unique_ptr<basic_node_type> node) noexcept
+            : line(line),
+              indent(indent),
+              state(state),
+              p_node(node.get()),
+              owned_node(std::move(node)) {
+        }
+
+        // Parse contexts are move-only so that the ownership of an owned node cannot be duplicated.
+        parse_context(const parse_context&) = delete;
+        parse_context& operator=(const parse_context&) = delete;
         parse_context(parse_context&&) noexcept = default;
         parse_context& operator=(parse_context&&) noexcept = default;
-
-        ~parse_context() {
-            switch (state) {
-            case context_state_t::BLOCK_MAPPING_EXPLICIT_KEY:
-            case context_state_t::FLOW_SEQUENCE_KEY:
-            case context_state_t::FLOW_MAPPING_KEY:
-                delete p_node;
-                p_node = nullptr;
-                break;
-            default:
-                break;
-            }
-        }
+        ~parse_context() = default;
 
         /// The current line. (count from zero)
         uint32_t line {0};
@@ -106,6 +114,8 @@ class basic_deserializer {
         context_state_t state {context_state_t::BLOCK_MAPPING};
         /// The pointer to the associated node to this context.
         basic_node_type* p_node {nullptr};
+        /// The node owned by this context, if any. Empty if p_node is owned by the result tree.
+        std::unique_ptr<basic_node_type> owned_node {};
     };
 
     /// @brief Definitions of state types for expected flow token hints.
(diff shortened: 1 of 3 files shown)


sndth:
@AlexandrKhromov2005 this is the PR we discussed in #536 - the `unique_ptr` version, with the
regression cases you listed added to `test_fuzz_regression.cpp`. Would be glad to have your review!

@fktn-k I can't set the Reviewers field myself, that needs write access to the repository, so
please assign him formally if you'd like it recorded on the PR.

sndth:
Coveralls was right, and the uncovered line is worth explaining because it isn't new code.

Line 1389 is the empty stack guard in `current_context()` from #552. The inputs which covered it were
that PR's own regression cases, `{{{}},{` and `{{{}},[`. With this change they fail earlier, at
`No corresponding flow mapping beginning is found.`, because a closing flow collection is now treated
as a mapping key based on the context actually owning its node rather than on `state`, so the stack no
longer empties there. I swept 27 million short inputs over two alphabets looking for anything which
still reaches the throw and found none, where the same sweep produced 8 hits before the change.

I marked it with `LCOV_EXCL_START/STOP`, the same way the identical guard in `pop_to_parent_node()` is
marked in this file, and noted in the comment why it stays. I'd rather keep the check than remove
it: it protects five call sites from dereferencing an invalid iterator, and "no input found in 27
million" is not a proof of unreachability.

If you'd prefer it removed instead, say so and I will drop it in a follow-up rather than in this PR.

fktn-k:
@sndth  
I didn't know GitHub requires a reviewer to be a collabolator.  

@AlexandrKhromov2005   
Could you review the PR and approve it if there is no issue found?  
You should be able to perform an approval.

tests/unit_test/test_fuzz_regression.cpp
```diff
@@ -154,6 +154,32 @@ TEST_CASE("FuzzRegression") {
         REQUIRE_THROWS_AS(root = fkyaml::node::deserialize(p_begin, p_end), fkyaml::parse_error);
     }
 
+    SUBCASE("key nodes owned by a parse context are released") {
+        // These inputs leave a key node owned by its parse context on a path where the context state is
+        // later rewritten, which used to leak the node. They are kept here so the sanitizer builds cover
+        // that ownership, the parse results themselves are secondary.
+        auto input = GENERATE(std::string("{{{"), std::string("? ["));
+        REQUIRE_NOTHROW(root = fkyaml::node::deserialize(input));
```
fktn-k, line 162: (This is an announcement, not a review.)  
I created an issue (#564) for the wrong parse results.

include/fkYAML/detail/input/deserializer.hpp
```diff
@@ -929,16 +940,17 @@ class basic_deserializer {
...
+                std::unique_ptr<basic_node_type> owned_node = std::move(last_context.owned_node);
                 m_context_stack.pop_back();
 
-                // handle cases where the flow mapping is a mapping key node.
+                // handle cases where the flow mapping is a mapping key node. A context owns its node only
+                // while that node is a key which has not been added to its parent mapping yet.
 
-                if (!m_context_stack.empty() && state == context_state_t::FLOW_MAPPING_KEY) {
+                if (!m_context_stack.empty() && owned_node != nullptr) {
                     basic_node_type key_node = std::move(*mp_current_node);
```
fktn-k: Wouldn't this convey the intended meaning more clearly?  

```suggestion
                    basic_node_type key_node = std::move(owned_node);
```

This also applies to the same part in the `lexical_token_t::SEQUENCE_FLOW_END` handling.

fktn-k (changes requested):

include/fkYAML/detail/input/deserializer.hpp
sndth: Yes! This convey would be more clearly.

fktn-k (approved):
LGTM

AlexandrKhromov2005 (approved):
Reviewed the full diff — this is the right fix, approving.

The ownership is now held by `owned_node` (a `std::unique_ptr`) instead of being re-derived from the mutable `state`, and making `parse_context` move-only is what actually closes the hole: the copy that used to duplicate ownership (and turned my `owns_node` prototype into a use-after-free) is now a compile error rather than a latent double-free.

The part I looked at hardest is the flow-end handling in the `*_FLOW_END` cases. Moving `owned_node` out of the context *before* `pop_back()` is the correct ordering: the node outlives the pop, and is then either moved into `key_node` and grafted, or dropped by the local `unique_ptr`. Both the empty-stack path and the borrowed-node path (`owned_node == nullptr`) do the right thing, so there's no leak and no dangling `mp_current_node` use.

The regression subcases cover the family well — `{{{`, `? [`, `? []`, and especially `? a / : / - 1`, which is the shape that exposed the UAF when ownership was a flag. The `current_context` LCOV_EXCL note is reasonable: with the key node owned by its context, the inputs that used to reach that guard now fail earlier.

Thanks @sndth for driving this to the clean solution.

fktn-k:
@sndth @AlexandrKhromov2005  
Thanks for the contribution!  
I've just merged this into develop.

PR #567, opened 2026-08-20
Improve yaml test suite workflow

fktn-k:
This PR is a follow-up of #562 and [the feedbacks](https://github.com/fktn-k/fkYAML/pull/562#issuecomment-5354366266) from @sndth.  
The following changes have been made to improve the YAML test suite workflow.  
- Optimized the YAML test suite CI flow by adding a baseline-comparison pipeline with cache reuse, so repeated runs can avoid rebuilding/retesting the same baseline commit.
- Updated failure parsing so "Skipped" is treated as a successful status (not a failure).
  - Although there is no "Skipped" test cases now, this should be a guard for future changes.
- Hardened the gate by failing loudly on unrecognized CTest progress/status lines, so malformed or new output formats cannot silently bypass new-failure detection.

---

## Pull Request Checklist

## Please don't

### DIFF
--- a/.github/workflows/yaml_test_suite.yml
+++ b/.github/workflows/yaml_test_suite.yml
@@ -62,7 +62,7 @@ jobs:
         key: yaml-test-suite-${{ env.YAML_TEST_SUITE_TAG }}
 
     - name: Configure CMake
-      run: cmake -B ${{github.workspace}}/build -DCMAKE_BUILD_TYPE=Debug -DFK_YAML_USE_YAML_TEST_SUITE=ON -DFETCHCONTENT_SOURCE_DIR-YAML-TEST-SUITE=${{env.YAML_TEST_SUITE_DIR}}
+      run: cmake -B ${{github.workspace}}/build -DCMAKE_BUILD_TYPE=Debug -DFK_YAML_USE_YAML_TEST_SUITE=ON -DFETCHCONTENT_SOURCE_DIR_YAML-TEST-SUITE=${{env.YAML_TEST_SUITE_DIR}}
 
     - name: Build
       run: cmake --build ${{github.workspace}}/build --config Debug -j ${{env.JOBS}}
@@ -146,7 +146,7 @@ jobs:
 
     - name: Configure CMake (baseline)
       if: steps.test.outcome == 'failure' && steps.baseline_ref.outputs.skip != 'true' && steps.baseline_cache_restore.outputs.cache-hit != 'true'
-      run: cmake -B ${{github.workspace}}/baseline/build -S ${{github.workspace}}/baseline -DCMAKE_BUILD_TYPE=Debug -DFK_YAML_USE_YAML_TEST_SUITE=ON -DFETCHCONTENT_SOURCE_DIR-YAML-TEST-SUITE=${{env.YAML_TEST_SUITE_DIR}}
+      run: cmake -B ${{github.workspace}}/baseline/build -S ${{github.workspace}}/baseline -DCMAKE_BUILD_TYPE=Debug -DFK_YAML_USE_YAML_TEST_SUITE=ON -DFETCHCONTENT_SOURCE_DIR_YAML-TEST-SUITE=${{env.YAML_TEST_SUITE_DIR}}
 
     - name: Build (baseline)
       if: steps.test.outcome == 'failure' && steps.baseline_ref.outputs.skip != 'true' && steps.baseline_cache_restore.outputs.cache-hit != 'true'
--- a/.github/workflows/yaml_test_suite.yml
+++ b/.github/workflows/yaml_test_suite.yml
@@ -67,13 +67,16 @@ jobs:
     - name: Build
       run: cmake --build ${{github.workspace}}/build --config Debug -j ${{env.JOBS}}
 
+    - name: Test validation compositor
+      run: ctest --test-dir ${{github.workspace}}/build -C Debug --output-on-failure -j -R 'validation_compositor_' ${{env.JOBS}}
+
     - name: Test
       id: test
       # Save the test results to a log file to avoid duplicate ctest runs in the next step.
       # The `pipefail` option ensures that the `ctest` exit code is propagated through the `tee` command.
       run: |
         set -o pipefail
-        ctest --test-dir ${{github.workspace}}/build -C Debug --output-on-failure -j ${{env.JOBS}} \
+        ctest --test-dir ${{github.workspace}}/build -C Debug --output-on-failure -R 'yaml_test_suite_' -j ${{env.JOBS}} \
           | tee ${{github.workspace}}/current_ctest.log
       # This allows the workflow to continue even if some tests fail, so we can generate a report of the failures.
       continue-on-error: true
@@ -172,7 +175,7 @@ jobs:
       # The `pipefail` option ensures that the `ctest` exit code is propagated through the `tee` command.
       run: |
         set -o pipefail
-        ctest --test-dir ${{github.workspace}}/baseline/build -C Debug --output-on-failure -j ${{env.JOBS}} \
+        ctest --test-dir ${{github.workspace}}/baseline/build -C Debug --output-on-failure -R 'yaml_test_suite_' -j ${{env.JOBS}} \
           | tee ${{github.workspace}}/baseline_ctest.log
       continue-on-error: true
 
--- a/.github/workflows/yaml_test_suite.yml
+++ b/.github/workflows/yaml_test_suite.yml
@@ -68,7 +68,7 @@ jobs:
       run: cmake --build ${{github.workspace}}/build --config Debug -j ${{env.JOBS}}
 
     - name: Test validation compositor
-      run: ctest --test-dir ${{github.workspace}}/build -C Debug --output-on-failure -j -R 'validation_compositor_' ${{env.JOBS}}
+      run: ctest --test-dir ${{github.workspace}}/build -C Debug --output-on-failure -R 'validation_compositor_' -j ${{env.JOBS}}
 
     - name: Test
       id: test
--- a/.github/workflows/yaml_test_suite.yml
+++ b/.github/workflows/yaml_test_suite.yml
@@ -16,6 +16,12 @@ on:
       - CMakeLists.txt
   workflow_dispatch:
 
+concurrency:
+  # Cancel a superseded run on the same branch/PR so we don't waste CI time fully
+  # building/testing a commit that's no longer the latest.
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
 jobs:
   yaml-test-suite:
     # Builds and tests both the current ref and a baseline ref (PR base branch, or the
@@ -87,8 +93,20 @@ jobs:
           echo "skip=true" >> "$GITHUB_OUTPUT"
         fi
 
-    - name: Checkout baseline
+    - name: Restore baseline cache
+      # Baseline results only depend on the baseline commit and this report script, so cache them
+      # by ref to skip re-building/re-testing an already-evaluated baseline on repeated runs. Uses
+      # the split restore/save actions (instead of the combined action) because this same
+      # `baseline_cache` directory is later reused to cache this run's own commit for future PRs.
+      id: baseline_cache_restore
       if: steps.test.outcome == 'failure' && steps.baseline_ref.outputs.skip != 'true'
+      uses: actions/cache/restore@v4
+      with:
+        path: baseline_cache
+        key: yaml-test-suite-baseline-${{ steps.baseline_ref.outputs.ref }}-${{ hashFiles('scripts/report_yaml_test_suite_failures.py') }}
+
+    - name: Checkout baseline
+      if: steps.test.outcome == 'failure' && steps.baseline_ref.outputs.skip != 'true' && steps.baseline_cache_restore.outputs.cache-hit != 'true'
       uses: actions/checkout@v7
       with:
         ref: ${{ steps.baseline_ref.outputs.ref }}
@@ -96,18 +114,18 @@ jobs:
         fetch-depth: 0
 
     - name: Configure CMake (baseline)
-      if: steps.test.outcome == 'failure' && steps.baseline_ref.outputs.skip != 'true'
+      if: steps.test.outcome == 'failure' && steps.baseline_ref.outputs.skip != 'true' && steps.baseline_cache_restore.outputs.cache-hit != 'true'
       run: cmake -B ${{github.workspace}}/baseline/build -S ${{github.workspace}}/baseline -DCMAKE_BUILD_TYPE=Debug -DFK_YAML_USE_YAML_TEST_SUITE=ON
 
     - name: Build (baseline)
-      if: steps.test.outcome == 'failure' && steps.baseline_ref.outputs.skip != 'true'
+      if: steps.test.outcome == 'failure' && steps.baseline_ref.outputs.skip != 'true' && steps.baseline_cache_restore.outputs.cache-hit != 'true'
       run: cmake --build ${{github.workspace}}/baseline/build --config Debug -j ${{env.JOBS}}
 
     - name: Check baseline has yaml_test_suite tests
       # A baseline ref predating the yaml_test_suite integration has no matching tests at all, so
       # every current failure would incorrectly look "new"; detect that and skip the comparison.
       id: baseline_availability
-      if: steps.test.outcome == 'failure' && steps.baseline_ref.outputs.skip != 'true'
+      if: steps.test.outcome == 'failure' && steps.baseline_ref.outputs.skip != 'true' && steps.baseline_cache_restore.outputs.cache-hit != 'true'
       run: |
         if ctest --test-dir ${{github.workspace}}/baseline/build -N -R 'yaml_test_suite_' 2>&1 | grep -q "Total Tests: 0"; then
           echo "Baseline (${{ steps.baseline_ref.outputs.ref }}) has no yaml_test_suite tests; skipping new-failure comparison."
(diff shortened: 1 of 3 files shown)


fktn-k:
@sndth  
I've addressed the feedback in #562.  
Could you review the changes if you have time?

sndth:
Of course! I ran a synthetic ctest project that produces every status and fed the real output
through the script:

**The `Skipped` change is a no-op (it never matches)**: ctest prefixes every non-passed status with
`***`, so the captured status is `***Skipped`, not `Skipped`:

```
1/5 Test #1: yaml_test_suite_AAAA_YAML ........   Passed    0.02 sec
3/5 Test #3: yaml_test_suite_CCCC_YAML ........***Skipped   0.02 sec
```

There's a second, independent path too: the summary section prints:

```
The following tests did not run:
	  3 - yaml_test_suite_CCCC_YAML (Skipped)
```

and the `SUMMARY_LINE_RE` branch appends a `FailureRecord` without looking at `group("status")` at
all. So even with the first one fixed, a skipped case still lands in `--names-output` and trips the
ratchet. `DISABLED TRUE` behaves the same way (`***Not Run (Disabled)` / `(Disabled)`). CI is green
here only because nothing is skipped today, so the guard doesn't guard yet.

What I verified works, on both a normal and a `-j 2` log:

```python
# ctest prefixes every non-passed status with "***" (e.g. "***Skipped", "***Not Run").
SUCCESS_STATUSES = {"Passed", "Skipped", "Disabled", "Not Run (Disabled)"}

def normalize_status(status: str) -> str:
    return status.strip().lstrip("*").strip()
```

applied in both branches: `not in SUCCESS_STATUSES` on the progress line, and an early `continue`
on the summary line.

**The loud failure and the cache design LGTM.**

Two things they don't reach, though, both about the cold path that the 20-minute timeout actually
threatens:

- `FETCHCONTENT_BASE_DIR` wouldn't have helped anyway: `FetchContent_Populate(yaml-test-suite
  GIT_REPOSITORY ...)` is the multi-argument form, which ignores it. An `actions/cache` entry on
  `yaml-test-suite-src` is the remaining option.
- ctest runs **twice per tree**: once in `Test`, then again inside the script via `run_ctest()`.
  Same for the baseline. The script already has `--log-file`, so a `ctest -R yaml_test_suite_... |
  tee ctest.log` step feeding `--log-file ctest.log` removes two full runs.

sndth:
I'm happy to send the status normalisation as a PR against this branch, or to add a few
`parse_failures()` unit tests over canned ctest logs so this can't regress silently 😸

fktn-k:
@sndth  
Thank you for the feedbacks!  

> The Skipped change is a no-op (it never matches)

I should have noticed it. Since you've already fix it, could you upload the patch as a branch in your fork? I'd like to merge it into this branch. I think it should be fine if you push a branch in your fork and I `cherry-pick` it to this branch.

> `FETCHCONTENT_BASE_DIR` wouldn't have helped anyway: `FetchContent_Populate(yaml-test-suite GIT_REPOSITORY ...)` is the multi-argument form, which ignores it. An actions/cache entry on `yaml-test-suite-src` is the remaining option.

Based on your feedback, I explored ways to avoid the duplicate installations.  
I think the best approach would be to define a new CMake variable, `FK_YAML_YAML_TEST_SUITE_DIR`: if some path is set to the variable, then CMake will use that path and skip installing the test suite. In the CI workflow, the test suite is manually downloaded once and the path is set to the variable. Caching it should further improve efficiency.  
I'll try to implement it. Let me know if you have better ideas.


> ctest runs twice per tree: once in Test, then again inside the script via run_ctest().
Same for the baseline. The script already has --log-file, so a ctest -R yaml_test_suite_... | tee ctest.log step feeding --log-file ctest.log removes two full runs.

Your suggestion sounds great! I'll adjust the workflow accordingly.

sndth:
> I should have noticed it. Since you've already fix it, could you upload the patch as a branch in your fork? I'd like to merge it into this branch. I think it should be fine if you push a branch in your fork and I `cherry-pick` it to this branch.

Pushed as [fix_skipped_status_detection](https://github.com/sndth/fkYAML/tree/fix_skipped_status_detection), branched off this PR's head so it should cherry-picks cleanly.

`Skipped` and `Not Run (Disabled)` are now excluded on both the progress line and the summary line -
the summary branch was appending a record without looking at the status at all. Verified on synthetic
ctest logs covering every status; against a real run it reports the same `124` failing cases as before,
so nothing changes today.

> I think the best approach would be to define a new CMake variable

On the test suite download: rather than a new variable, switching the runner from the multi-argument
`FetchContent_Populate(yaml-test-suite GIT_REPOSITORY ...)` to `FetchContent_Declare` +
`FetchContent_MakeAvailable` may be simpler. That form honours the built-in
`FETCHCONTENT_SOURCE_DIR_YAML-TEST-SUITE`, so CI can point at a pre-downloaded, cached copy without a
project-specific variable, and local builds get the same. `tests/fuzz_test/CMakeLists.txt` already
uses that pattern on a data-only repository, so it works here - `MakeAvailable` only calls
`add_subdirectory` when the populated directory contains a `CMakeLists.txt`.

fktn-k:
@sndth  
> On the test suite download: rather than a new variable, switching the runner from the multi-argument `FetchContent_Populate(yaml-test-suite GIT_REPOSITORY ...)` to `FetchContent_Declare` + `FetchContent_MakeAvailable` may be simpler. That form honours the built-in `FETCHCONTENT_SOURCE_DIR_YAML-TEST-SUITE`, so CI can point at a pre-downloaded, cached copy without a project-specific variable, and local builds get the same. `tests/fuzz_test/CMakeLists.txt` already uses that pattern on a data-only repository, so it works here - `MakeAvailable` only calls `add_subdirectory` when the populated directory contains a `CMakeLists.txt`.

Oh, I misunderstood the comment on `FetchContent_Populate`. Thank you for clarifying.  
The suggested approach (`FetchContent_Declare` + `FetchContent_MakeAvailable`) would be much simpler. I'll switch to that approach.

fktn-k:
@sndth  
Done! I've made the following changes. I'd appreciate it if you could review them.

- **Treat skipped and disabled CTest statuses as successfull**
  - I `cherry-pick`ed your fix.
- **Reuse the YAML test suite**
  - If the YAML test suite is not cached, it will be downloaded and cached at the beginning of the workflow. Since the test data is pinned with a tag, the cached data will generally be reused without needing to be downloaded.
  - `FetchContent_Populate` is replaced with `FetchContent_Declare` and `FetchContent_MakeAvailable`. When configuring CMake, the test data directory is specified using the builtin variable `FETCHCONTENT_SOURCE_DIR_YAML-TEST-SUITE`. This prevents the data from being downloaded.
- **Avoid duplicate CTest runs**
  - Each CTest execution log is saved to a file. By specifying it to the script with the `log-file` option, unnecessary CTest executions within the script are eliminated.

.github/workflows/yaml_test_suite.yml
```diff
@@ -26,19 +32,49 @@ jobs:
...
+    - name: Save YAML Test Suite cache
+      if: steps.yaml_test_suite_cache_restore.outputs.cache-hit != 'true'
+      uses: actions/cache/save@v4
+      with:
+        path: ${{env.YAML_TEST_SUITE_DIR}}
+        key: yaml-test-suite-${{ env.YAML_TEST_SUITE_TAG }}
+
     - name: Configure CMake
-      run: cmake -B ${{github.workspace}}/build -DCMAKE_BUILD_TYPE=Debug -DFK_YAML_USE_YAML_TEST_SUITE=ON
+      run: cmake -B ${{github.workspace}}/build -DCMAKE_BUILD_TYPE=Debug -DFK_YAML_USE_YAML_TEST_SUITE=ON -DFETCHCONTENT_SOURCE_DIR-YAML-TEST-SUITE=${{env.YAML_TEST_SUITE_DIR}}
```
sndth: This needs an underscore. The built-in variable is `FETCHCONTENT_SOURCE_DIR_<uppercaseName>`, so for a
dependency named `yaml-test-suite` it has to be `FETCHCONTENT_SOURCE_DIR_YAML-TEST-SUITE`. You wrote it
correctly in the comment above, so this looks like a slip in the YAML.

I configured both ways against a local copy of the suite:

| variable | result |
|---|---|
| `...SOURCE_DIR-YAML-TEST-SUITE` | `CMake Warning (unused-cli): Manually-specified variables were not used`, and the suite is cloned anyway |
| `...SOURCE_DIR_YAML-TEST-SUITE` | nothing downloaded, 681 test cases generated, configure in 6s |

So the cache is populated and restored, and CMake still clones on every run. Everything stays green,
the only signal is one warning line in the configure output. Line 149 needs the same fix.

```suggestion
      run: cmake -B ${{github.workspace}}/build -DCMAKE_BUILD_TYPE=Debug -DFK_YAML_USE_YAML_TEST_SUITE=ON -DFETCHCONTENT_SOURCE_DIR_YAML-TEST-SUITE=${{env.YAML_TEST_SUITE_DIR}}
```

.github/workflows/yaml_test_suite.yml
```diff
@@ -87,27 +124,39 @@ jobs:
...
       with:
         ref: ${{ steps.baseline_ref.outputs.ref }}
         path: baseline
         fetch-depth: 0
 
     - name: Configure CMake (baseline)
-      if: steps.test.outcome == 'failure' && steps.baseline_ref.outputs.skip != 'true'
-      run: cmake -B ${{github.workspace}}/baseline/build -S ${{github.workspace}}/baseline -DCMAKE_BUILD_TYPE=Debug -DFK_YAML_USE_YAML_TEST_SUITE=ON
+      if: steps.test.outcome == 'failure' && steps.baseline_ref.outputs.skip != 'true' && steps.baseline_cache_restore.outputs.cache-hit != 'true'
+      run: cmake -B ${{github.workspace}}/baseline/build -S ${{github.workspace}}/baseline -DCMAKE_BUILD_TYPE=Debug -DFK_YAML_USE_YAML_TEST_SUITE=ON -DFETCHCONTENT_SOURCE_DIR-YAML-TEST-SUITE=${{env.YAML_TEST_SUITE_DIR}}
```
sndth: Same underscore fix as above.

```suggestion
      run: cmake -B ${{github.workspace}}/baseline/build -S ${{github.workspace}}/baseline -DCMAKE_BUILD_TYPE=Debug -DFK_YAML_USE_YAML_TEST_SUITE=ON -DFETCHCONTENT_SOURCE_DIR_YAML-TEST-SUITE=${{env.YAML_TEST_SUITE_DIR}}
```

.github/workflows/yaml_test_suite.yml
```diff
@@ -56,6 +92,7 @@ jobs:
       run: |
         echo "Generating YAML Test Suite Failures Report..."
         python3 ${{github.workspace}}/scripts/report_yaml_test_suite_failures.py --test-dir build \
+          --log-file ${{github.workspace}}/current_ctest.log \
           --output ${{github.workspace}}/${{steps.artifact_naming.outputs.artifact_name}} \
           --names-output ${{github.workspace}}/current_failures.txt
```
sndth, line 101: With the suite now outside the build tree, `detect_yaml_test_suite_root()` finds neither of the two
directories it probes:

```
<build>/tests/yaml_test_suite_runner/yaml-test-suite-src   missing
<build>/_deps/yaml-test-suite-src                          missing
```

Running the script the way this step now does gives 230 `(not found)` entries and `(empty input)` for
all 115 failing cases, so the per-case YAML section of the report comes out blank. With the root passed
explicitly there are no `(not found)` entries at all.

```suggestion
        python3 ${{github.workspace}}/scripts/report_yaml_test_suite_failures.py --test-dir build \
          --log-file ${{github.workspace}}/current_ctest.log \
          --yaml-test-suite-root ${{env.YAML_TEST_SUITE_DIR}} \
          --output ${{github.workspace}}/${{steps.artifact_naming.outputs.artifact_name}} \
          --names-output ${{github.workspace}}/current_failures.txt
```

.github/workflows/yaml_test_suite.yml
```diff
@@ -117,18 +166,60 @@ jobs:
...
+          | tee ${{github.workspace}}/baseline_ctest.log
       continue-on-error: true
 
     - name: Report Failures (baseline)
-      if: steps.test.outcome == 'failure' && steps.baseline_ref.outputs.skip != 'true' && steps.baseline_availability.outputs.present == 'true'
+      if: steps.test.outcome == 'failure' && steps.baseline_ref.outputs.skip != 'true' && steps.baseline_cache_restore.outputs.cache-hit != 'true' && steps.baseline_availability.outputs.present == 'true'
       run: |
         python3 ${{github.workspace}}/scripts/report_yaml_test_suite_failures.py --workspace ${{github.workspace}}/baseline --test-dir build \
+          --log-file ${{github.workspace}}/baseline_ctest.log \
           --names-output ${{github.workspace}}/baseline_failures.txt
```
sndth, line 188: Same here. This step only uses `--names-output`, so the missing input data has no effect on the ratchet
today - but passing the root keeps both invocations consistent, and avoids a surprise if this step ever
starts producing a report too.

```suggestion
        python3 ${{github.workspace}}/scripts/report_yaml_test_suite_failures.py --workspace ${{github.workspace}}/baseline --test-dir build \
          --log-file ${{github.workspace}}/baseline_ctest.log \
          --yaml-test-suite-root ${{env.YAML_TEST_SUITE_DIR}} \
          --names-output ${{github.workspace}}/baseline_failures.txt
```

.github/workflows/yaml_test_suite.yml
```diff
@@ -26,19 +32,49 @@ jobs:
...
       run: cmake --build ${{github.workspace}}/build --config Debug -j ${{env.JOBS}}
 
     - name: Test
       id: test
-      run: ctest --test-dir ${{github.workspace}}/build -C Debug --output-on-failure -j ${{env.JOBS}}
+      # Save the test results to a log file to avoid duplicate ctest runs in the next step.
+      # The `pipefail` option ensures that the `ctest` exit code is propagated through the `tee` command.
+      run: |
+        set -o pipefail
+        ctest --test-dir ${{github.workspace}}/build -C Debug --output-on-failure -j ${{env.JOBS}} \
```
sndth: The script used to run `ctest -R yaml_test_suite_` itself, so the log it parsed was filtered. This one
isn't:

```
without -R : Total Tests: 689
with    -R : Total Tests: 681
```

The extra eight are the `validation_compositor_*` runner tests, which now feed both the failure report
and the ratchet. That may well be what you want: if the harness itself breaks you would want the
workflow to say so - but it's a silent change of scope, and a failure there would produce a report row
with no input data behind it, since those tests have no suite case to point at.

Worth deciding deliberately either way. Adding `-R yaml_test_suite_` here keeps the previous behaviour.

sndth (changes requested):
Cherry-pick, `set -o pipefail` with `tee`, and switch to `FetchContent_Declare` + `FetchContent_MakeAvailable` LGTM.

Two of the comments below matter: without them the new caching doesn't take effect and the report
comes out with no input data. The third is a question rather than a defect.

.github/workflows/yaml_test_suite.yml
fktn-k: Sorry for pointing out such a simple mistake.  
I applied the suggestion.

.github/workflows/yaml_test_suite.yml
fktn-k: Similarly, I applied the suggestion.

.github/workflows/yaml_test_suite.yml
fktn-k: Thank you for pointing this out.  
You are right. The option should be passed to the script.  
I applied the suggestion.

.github/workflows/yaml_test_suite.yml
fktn-k: I applied the suggestion here too to align with the one for the target commit.

.github/workflows/yaml_test_suite.yml
fktn-k: I decided to run the runner tests in a separate step before running the cases for the YAML test suite. This is to ensure that failures are detected early if there are issues with the runner itself, given that the runner's implementation might change in the future.  
Then, I added `-R 'yaml_test_suite_'` for CTest executions for the YAML test suite.  
As a result, the CTest log for the YAML test suite's test cases now record the same total number of test cases as before.

fktn-k:
@sndth  I finished making the fixes.

.github/workflows/yaml_test_suite.yml
```diff
@@ -26,19 +32,52 @@ jobs:
...
+
     - name: Configure CMake
-      run: cmake -B ${{github.workspace}}/build -DCMAKE_BUILD_TYPE=Debug -DFK_YAML_USE_YAML_TEST_SUITE=ON
+      run: cmake -B ${{github.workspace}}/build -DCMAKE_BUILD_TYPE=Debug -DFK_YAML_USE_YAML_TEST_SUITE=ON -DFETCHCONTENT_SOURCE_DIR_YAML-TEST-SUITE=${{env.YAML_TEST_SUITE_DIR}}
 
     - name: Build
       run: cmake --build ${{github.workspace}}/build --config Debug -j ${{env.JOBS}}
 
+    - name: Test validation compositor
+      run: ctest --test-dir ${{github.workspace}}/build -C Debug --output-on-failure -j -R 'validation_compositor_' ${{env.JOBS}}
```
sndth: Small one: `-j` has no value here and `${{env.JOBS}}` ends up as a trailing positional, so the job
limit is not applied. It still runs - on ctest 4.4 a bare `-j` means unbounded and the stray `2` is
ignored - but a ctest that requires a value for `-j` would reject it.

```suggestion
      run: ctest --test-dir ${{github.workspace}}/build -C Debug --output-on-failure -R 'validation_compositor_' -j ${{env.JOBS}}
```

sndth (approved):
All looks good to me! I reproduced the whole chain locally: the suite is no longer downloaded, the
report has the input YAML back, and the compositor tests are out of the log.

Splitting them into their own step is better than either option I suggested. Thank you for taking my
proposals - I have one more small point below, but it really is just a detail 😄

.github/workflows/yaml_test_suite.yml
fktn-k: Done!

fktn-k:
@sndth Thank you for reviewing. I just fixed the order of ctest options which you pointed out in the last review. Since everything is resolved I'll merge this into develop!

PR #571, opened 2026-08-21
Support Asan/Fuzzer on MSVC

sndth:
Follow-up to #566, which picked up an overlooked review comment from #546: `tests/fuzz_test/CMakeLists.txt`
used the Clang spelling of the sanitizer flags, so the target could not be configured with MSVC even
though MSVC ships both Asan and libFuzzer.

The failure mode was worse than "won't configure": `cl` accepts `-fsanitize=fuzzer,address,undefined`
as three unknown options (warning D9002), drops them, and the build only falls over at link time with
`LNK1561: entry point must be defined`. Nothing tells you the sanitizers were never enabled.

### What's changed

- `tests/fuzz_test/CMakeLists.txt` grows an `if(MSVC)` branch using `/fsanitize=address /fsanitize=fuzzer`.
  The Clang branch is unchanged.
  - `/Zi` and `/DEBUG` are set in every configuration. Without them Asan warns (C5072, LNK4302) and
    C5072 is fatal under the shared test config's warnings-as-errors.
  - No optimization flag is set: the build type already picks one, and `/O1` conflicts with the `/RTC1`
    that the Debug configuration adds (D8016).
  - MSVC has no UBSan, so `undefined` is left out rather than passed and silently ignored.
- The MSVC Asan runtime is a DLL shipped next to `cl.exe`. Without it `fuzz_test.exe` cannot start at
  all (`STATUS_DLL_NOT_FOUND`), so it is copied next to the binary in a POST_BUILD step. The path is
  derived from `CMAKE_CXX_COMPILER`, which also picks the right architecture automatically.
- `run_fuzz_test` passes the dictionary with native separators on Windows. libFuzzer cannot open
  `-dict=` through a forward-slash path and carries on **without a dictionary** rather than failing,
  so this one is easy to lose silently. The corpus directories and `-artifact_prefix=` work either way
  and are untouched.
- Fixed a pre-existing dead branch in the top-level `CMakeLists.txt`:

  ```cmake
  if (CXX_COMPILER_ID STREQUAL "Clang")    # not a CMake variable, always empty -> always else
    ...
  endif()
  set(FK_YAML_BUILD_FUZZ_TEST ON)          # unconditionally overrode both branches
  ```

  `FK_YAML_BUILD_ALL_TEST` therefore always enabled the fuzz test and always printed the
  "only supported with Clang" warning. It now checks `CMAKE_CXX_COMPILER_ID` and accepts Clang or MSVC.

### Verification

Built from scratch with VS 2026 (MSVC 14.51): x32/64 (Debug/Release, all four
clean. `run_fuzz_test` loads the dictionary (74 entries), instruments 10320 coverage counters and exits
0, and the binary starts from a shell without the VS environment on `PATH`. A temporary
heap-buffer-overflow in `LLVMFuzzerTestOneInput` produced a symbolised Asan report and a crash artifact
in `corpus\findings\`, so the sanitizer is live rather than merely linked.

I couldn't build the Clang path locally. It is byte-identical to before, but worth flagging that CI
won't catch a mistake there either: no workflow builds the fuzz target, so `make fuzz-test` is the
only check that path gets.

---

## Pull Request Checklist

## Please don't


sndth:
I'd also add that nothing is wired into `windows.yml`. That workflow is a 16-job matrix on a 10-minute timeout and `run_fuzz_test` uses `-max_total_time=300`; a fuzzing job belongs somewhere opt-in, if anywhere.

@fktn-k
Please let me know what you're think about it.

fktn-k:
@sndth  
> I couldn't build the Clang path locally. It is byte-identical to before, but worth flagging that CI won't catch a mistake there either: no workflow builds the fuzz target, so make fuzz-test is the only check that path gets.

I confirmed that the Clang version works the same as before.  

> I'd also add that nothing is wired into windows.yml. That workflow is a 16-job matrix on a 10-minute timeout and run_fuzz_test uses -max_total_time=300; a fuzzing job belongs somewhere opt-in, if anywhere.
>
> @fktn-k
> Please let me know what you're think about it.

I think it should be better to set up some dedicated workflow for the fuzz tests other than platform-specific ones.

fktn-k (changes requested):
The changes look good to me.  
I confirmed that the fuzzing test can still be run with Clang 18 on Ubuntu 24.04, just as before.

Since I merged #572 first, could you resolve the conflicts?

sndth:
> The changes look good to me. I confirmed that the fuzzing test can still be run with Clang 18 on Ubuntu 24.04, just as before.
> 
> Since I merged #572 first, could you resolve the conflicts?

I just rebased onto the current `develop` to resolve this conflict.

fktn-k (approved):
LGTM

PR #572, opened 2026-08-21
Make UBSan findings fail the fuzz test

sndth:
UBSan checks are recoverable by default, so a finding in the fuzz target prints a `runtime error:`
line and the run continues. libFuzzer never sees a crash, never saves the offending input, and
`make fuzz-test` still exits 0. `-fno-sanitize-recover=all` makes those findings abort instead.

`signed-integer-overflow` is excluded to stay in step with the clang sanitizer flags for the unit
tests, which have excluded it since #391. `implicit-conversion` is excluded there too but is not part
of `-fsanitize=undefined`, so it isn't repeated here.

Only `target_compile_options` changes: recover mode is a codegen decision (the compiler emits
`__ubsan_handle_*_abort` rather than `__ubsan_handle_*`), so the flag is a no-op on the link line.

Two notes:

- I couldn't build this locally, and no workflow builds the fuzz target, so `make fuzz-test` on Linux
  is the first real check. The unit test job already runs a superset of these checks fatally with the
  same exclusion and is green, so the combination is at least viable on this codebase. If something
  does fire, it is most likely a finding rather than a regression of this change.
- This touches the same `target_compile_options` block as #571, so whichever merges second will need a
  trivial conflict resolution.

---

## Pull Request Checklist

## Please don't

### DIFF
--- a/tests/fuzz_test/CMakeLists.txt
+++ b/tests/fuzz_test/CMakeLists.txt
@@ -27,6 +27,11 @@ target_compile_options(
   fuzz_test
   PRIVATE
     -fsanitize=fuzzer,address,undefined
+    # UBSan checks are recoverable by default: without this they print a diagnostic and the
+    # run continues, so libFuzzer never sees a crash and never saves the offending input.
+    -fno-sanitize-recover=all
+    # Excluded for the same reason the unit test configuration excludes it.
+    -fno-sanitize=signed-integer-overflow
     -g
     -O1
 )


fktn-k:
Thank you for this PR!
I'm verifying this with Clang 18 on Ubuntu 24.04.

fktn-k (approved):
The changes look good to me!  
I didn't notice the misconfiguration.  
Also, the policy of aligning the sanitizer flags in unit tests is good.

Regarding the verification:

> Thank you for this PR!
> I'm verifying this with Clang 18 on Ubuntu 24.04.

I comfirmed that errors from UBSan was just printed and the fuzzer proceeded as if nothing happened on the develop HEAD, and that they correctly trigger a crash with the fix applied.

PR #573, opened 2026-08-21
Reject an input which leaves a flow collection unclosed

### LINKED ISSUE #564: An input with unclosed curly braces or square brackets should be treated as an error
### Description

Parsing malformed inputs like `{{{` or `? [` currently completes with no error and the output is also wrong.  
The parser should throw an exception against such an input.  

### Reproduction steps

Parse the following YAML with fkYAML.
```yaml
{{{
```

A `parse_error` is expected to be thrown, but the current implementation completes the parsing with no error and outputs a wrong deserialization result.

### Minimal code example

```Shell
#include <iostream>
#include <fkYAML/node.hpp>

int main() {
    try {
        auto root = fkyaml::node::deserialize("{{{");
　　　　 std::cout << root << std::endl;
    }
    catch (const fkyaml::parse_error& e) {
        // This path doesn't go through.
    }

}

// output:
// {}
```

### Error messages

```Shell

```

GCC 13.3.0 on Ubuntu 24.04

### Library version

0.4.3

### Validation

sndth:
Fixes #564.

`END_OF_DIRECTIVES` and `END_OF_DOCUMENT` already reject a document marker found inside a flow
collection, but nothing checked the flow context depth at the end of input, so `{{{` and `? []`
completed with a wrong result instead of an error. The `END_OF_BUFFER` case in the token switch only
handles an empty input: a non-empty document leaves the parse loop through its `while` condition,
so the check goes after the loop.

Note the change to `test_fuzz_regression.cpp`: that subcase was added in #563 to keep the sanitizer
builds covering the parse context ownership path, and it happened to pin the old behaviour with
`REQUIRE_NOTHROW`. Both of its inputs leave a flow collection unclosed, so they are now rejected. The
subcase keeps its original purpose, since the error is raised at the end of input, after the contexts
have been built and rewritten.

Unit tests pass, in the YAML test suite this also fixes `6JTT` (Flow sequence without closing bracket),
125 failing cases down to 124, with no newly failing case.

---

## Pull Request Checklist

## Please don't

### DIFF
--- a/include/fkYAML/detail/input/deserializer.hpp
+++ b/include/fkYAML/detail/input/deserializer.hpp
@@ -1160,6 +1160,10 @@ private:
             line = lexer.get_lines_processed();
         } while (token.type != lexical_token_t::END_OF_BUFFER);
 
+        if FK_YAML_UNLIKELY (m_flow_context_depth > 0) {
+            throw parse_error("An unclosed flow collection found at the end of input", line, indent);
+        }
+
         last_type = token.type;
     }
 
--- a/single_include/fkYAML/node.hpp
+++ b/single_include/fkYAML/node.hpp
@@ -8664,6 +8664,10 @@ private:
             line = lexer.get_lines_processed();
         } while (token.type != lexical_token_t::END_OF_BUFFER);
 
+        if FK_YAML_UNLIKELY (m_flow_context_depth > 0) {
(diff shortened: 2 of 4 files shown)


fktn-k (approved):
LGTM

fktn-k:
@sndth Thanks! I'll merge this into develop!

PR #574, opened 2026-08-21
Support single pair mappings in flow sequences

### LINKED ISSUE #561: Single pair mappings in a flow sequence are rejected
### Description

A flow sequence entry may be a single pair mapping written without braces, i.e. `[foo: 1]` means
`[{foo: 1}]`. fkYAML rejects that form. Writing the braces explicitly (`[{foo: 1}]`) works, and
nested flow collections in general are fine, so this is specific to the implicit single pair form.

This is the second of the two gaps split off from #557.

### Reproduction steps

Deserialize a flow sequence whose entry is a mapping entry without braces.

| input | expected | actual on `develop` (c9e7069) |
|---|---|---|
| `[foo: 1]` | `[{foo: 1}]` | `parse_error: No corresponding flow sequence beginning is found.` |
| `[foo: 1, bar: 2]` | `[{foo: 1}, {bar: 2}]` | `parse_error: No corresponding flow sequence beginning is found.` |
| `{a: [b: 1]}` | `{a: [{b: 1}]}` | `parse_error: No corresponding flow sequence beginning is found.` |
| `[{foo: 1}]` | `[{foo: 1}]` | correct today |

The expected column is what PyYAML 6.0.3 produces.

In the yaml-test-suite this affects `9MMW` (Single Pair Implicit Entries), which is rejected, and
`CFD4` (Empty implicit key in single pair flow sequences), which is worse: it parses but silently
drops the keys.

```yaml
- [ : empty key ]
- [: another empty key]
```

gives `[["empty key"], ["another empty key"]]` instead of a sequence of single pair mappings.
`CFD4` also needs empty keys from #557, so like `FRK4` it will only pass once both are addressed.

### Minimal code example

```cpp
#include <iostream>
#include <fkYAML/node.hpp>

int main() {
  // rejected
  fkyaml::node::deserialize(std::string("[foo: 1]"));

  // accepted, and this is what the input above should produce
  std::cout << fkyaml::node::serialize(fkyaml::node::deserialize(std::string("[{foo: 1}]")));
}
```

### Error messages

```Shell

```

MSVC; Windows 11

### Library version

v0.4.3

### Validation

sndth:
Fixes #561.

The wrapping itself was already there: when a key separator is found while the current node is a
sequence, `add_new_key` wraps the entry in a mapping and pushes a context for it. Nothing ever closed
that context, so `]` saw a block mapping on top of the stack instead of the flow sequence and reported
`No corresponding flow sequence beginning is found`. The comma had the same problem from the other
side (it did not close the wrapper either, so `[foo: 1, bar: 2]` would have collected both keys into
one mapping rather than two).

`close_single_pair_mapping()` is called at both points. Block content cannot appear inside a flow
collection, so a block mapping context found there can only be that wrapper.

I verified YAML test suite: 125 failing cases down to 116, with no newly failing case. Nine cases fixed:
`9MMW`, `CFD4`, `4FJ6`, `CN3R` and Spec Examples 7.4 (`LQZ7`), 7.8 (`87E4`), 7.11 (`L9U5`),
7.14 (`8UDB`) and 7.19 (`QF4Y`).

`CFD4` needed the empty key support from #558 as well, which is why it passes in full now.

Unit tests also pass.

---

## Pull Request Checklist

## Please don't

### DIFF
--- a/include/fkYAML/detail/input/deserializer.hpp
+++ b/include/fkYAML/detail/input/deserializer.hpp
@@ -882,6 +882,8 @@ private:
                     lexer.set_context_state(false);
                 }
 
+                close_single_pair_mapping(line, indent);
+
                 const bool has_valid_beginning =
                     !m_context_stack.empty() && (m_context_stack.back().state == context_state_t::FLOW_SEQUENCE ||
                                                  m_context_stack.back().state == context_state_t::FLOW_SEQUENCE_KEY);
@@ -1063,6 +1065,7 @@ private:
                 if FK_YAML_UNLIKELY (m_flow_token_state != flow_token_state_t::NEEDS_SEPARATOR_OR_SUFFIX) {
                     throw parse_error("invalid value separator is found.", line, indent);
                 }
+                close_single_pair_mapping(line, indent);
                 m_flow_token_state = flow_token_state_t::NEEDS_VALUE_OR_SUFFIX;
                 break;
             case lexical_token_t::ALIAS_PREFIX: {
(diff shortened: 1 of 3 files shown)


include/fkYAML/detail/input/deserializer.hpp
```diff
@@ -1583,6 +1586,28 @@ class basic_deserializer {
...
+    /// Block content cannot appear inside a flow collection, so a block mapping context found here can
+    /// only be that wrapper.
+    /// @param line Current line.
+    /// @param indent Current indentation.
+    void close_single_pair_mapping(const uint32_t line, const uint32_t indent) {
+        // LCOV_EXCL_START
+        if FK_YAML_UNLIKELY (m_context_stack.empty()) {
+            return;
+        }
+        // LCOV_EXCL_STOP
```
fktn-k, line 1607: Would it be better to throw a `parse_error` instead of silently `return`ing since such a condition should be unreachable?  
Excluding the code from coverage with `LCOV_EXCL_START` and `LCOV_EXCL_STOP` is fine.

fktn-k (changes requested):

include/fkYAML/detail/input/deserializer.hpp
sndth: You're right! I accidentally forgot to add `parse_error` there.

sndth:
One correction to the numbers above: since #573 landed, the baseline is 124 rather than 125,
so this is 124 -> 115. The nine fixed cases are unchanged, `6JTT` is not among them.

fktn-k (approved):
LGTM. Thanks a lot!

PR #575, opened 2026-08-22
Support an omitted value in a flow mapping entry

### LINKED ISSUE #560: An omitted value in a flow mapping is rejected
### Description

A mapping entry in a flow mapping may leave its value empty, in which case the value is null. fkYAML
rejects every form of it, with three different errors depending on what follows the `:`.

This is the first of the two gaps split off from #557 as you suggested. It is unrelated to empty
keys: the key here is present, only the value is missing.

### Reproduction steps

Deserialize a flow mapping whose entry has no value.

| input | expected | actual on `develop` (c9e7069) |
|---|---|---|
| `{foo: }` | `{foo: null}` | `parse_error: No corresponding flow mapping beginning is found.` |
| `{foo:}` | `{foo: null}` | `parse_error: The ":" mapping value indicator must be followed after a mapping key.` |
| `{foo: , bar: 1}` | `{foo: null, bar: 1}` | `parse_error: invalid value separator is found.` |
| `{foo: 1, bar: }` | `{foo: 1, bar: null}` | `parse_error: No corresponding flow mapping beginning is found.` |
| `{"foo": }` | `{foo: null}` | `parse_error: No corresponding flow mapping beginning is found.` |

The expected column is what PyYAML 6.0.3 produces.

In the yaml-test-suite this affects `4ABK` (Flow Mapping Separate Values), which contains an
`omitted value:,` entry, and `FRK4` (**Spec Example 7.3**, Completely Empty Flow Nodes). `FRK4` also
needs empty keys from #557, so it will only pass once both are done.

### Minimal code example

```cpp
#include <iostream>
#include <fkYAML/node.hpp>

int main() {
  // rejected
  fkyaml::node::deserialize(std::string("{foo: }"));

  // the block equivalent is accepted and yields {foo: null}
  std::cout << fkyaml::node::serialize(fkyaml::node::deserialize(std::string("foo:\n")));
}
```

### Error messages

```Shell

```

MSVC; Windows 11

### Library version

v0.4.3

### Validation

sndth:
Fixes #560.

Two layers were involved.

In the deserializer, a flow mapping entry whose value never arrives leaves its `MAPPING_VALUE`
context on the stack, and `}` requires the flow context to be on top, hence
`No corresponding flow mapping beginning is found`. The value node is already null, since
`add_new_key` default constructs it, so the entry only needs its context closed. That happens at both
`,` and `}`; at the separator it has to run before the token state check, otherwise
`{foo: , bar: 1}` is rejected first.

In the lexer, `{foo:}` is a separate problem: `determine_plain_scalar_range()` ended a plain scalar at
`:` only before a space, tab or newline, so `foo:` was scanned as one scalar. A `:` followed by a flow
indicator is not `ns-plain-safe` in a flow context, so it now ends the scalar there too, using the
same set of indicators the lexer already checks for a standalone colon.

I verified: unit tests pass, YAML test suite: 115 failing cases down to 113, fixing
`C2DT` (Spec Example 7.18) and `NKF9`, with no newly failing case. I also checked that empty
entries are still rejected: `{foo: 1,, bar: 2}`, `{,}`, `[a,,b]` and `[,]` all still throw.

Two cases named in the issue still fail, for reasons outside this change. `FRK4` also needs an
*explicit* key with an omitted value (`? foo :,`), which is a different context state. `4ABK` fails on
`http://foo.com` - a flow mapping entry with no value indicator at all, which should be
`{http://foo.com: null}`. `{a, b}` reproduces that on its own, and it is a separate gap from this one.

---

## Pull Request Checklist

## Please don't

### DIFF
--- a/include/fkYAML/detail/input/deserializer.hpp
+++ b/include/fkYAML/detail/input/deserializer.hpp
@@ -1006,6 +1006,8 @@ private:
                     lexer.set_context_state(false);
                 }
 
+                close_omitted_mapping_value(line, indent);
+
                 const bool has_valid_beginning =
                     !m_context_stack.empty() && (m_context_stack.back().state == context_state_t::FLOW_MAPPING ||
                                                  m_context_stack.back().state == context_state_t::FLOW_MAPPING_KEY);
@@ -1062,6 +1064,7 @@ private:
                 if FK_YAML_UNLIKELY (m_flow_context_depth == 0) {
                     throw parse_error("invalid value separator is found.", line, indent);
                 }
+                close_omitted_mapping_value(line, indent);
                 if FK_YAML_UNLIKELY (m_flow_token_state != flow_token_state_t::NEEDS_SEPARATOR_OR_SUFFIX) {
                     throw parse_error("invalid value separator is found.", line, indent);
                 }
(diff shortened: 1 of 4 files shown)


fktn-k (approved):
LGTM

fktn-k:
@sndth  
I'll merge this one first. Thanks!

PR #577, opened 2026-08-22
Raise a parse error when a collection tag is applied to a scalar

### LINKED ISSUE #576: A `!!map` tag on a block collection aborts the process
### Description

`deserialize()` aborts the process instead of reporting an error when a `!!map` tag ends up applying
to a scalar node. `FK_YAML_ASSERT(tag_type != tag_t::SEQUENCE && tag_type != tag_t::MAPPING)` in
`scalar_parser.hpp:75` fires, so a Debug build of an application dies on malformed input rather than
throwing `fkyaml::parse_error`.

The equivalent `!!seq` shapes are handled correctly, which is what makes this easy to miss:

| input | result |
|---|---|
| `foo: !!seq` + block sequence | parses |
| `- !!seq` + nested sequence | parses |
| `key: &a` / ` !!seq` / `- x` | parses |
| `- !!map` + block mapping | **aborts** |
| `key: &a` / ` !!map` / `a: b` | **aborts** |

This is not only reachable from spec constructs. Running the fuzz target with Asan found it from a
random mutation in under 800 executions, on this input:

```
tey: !!map line1- \r line2\r\n
```

```
#7  wassert
#8  scalar_parser<...>::parse_flow          scalar_parser.hpp:75
#9  basic_deserializer<...>::deserialize_node   deserializer.hpp:1116
```

This affects four YAML test suite cases, all aborting today: `6JWB`, `735Y`, `BU8L` and `H7J7`. Only
`H7J7` is marked as an error case; the other three are valid YAML that should parse, so raising a
`parse_error` stops the crash but does not close that gap.

### Reproduction steps

Build the snippet below in a Debug configuration, so `FK_YAML_ASSERT` is live, and run it.

Expected: a `fkyaml::parse_error`, or a successful parse applying the tag to the following block
collection.

Actual: the assertion fires and the process aborts with exit code 3. Building with `NDEBUG` skips the
assertion and the parse then fails with `Detected invalid indentation` instead, so the assertion is a
correct tripwire rather than an over-strict one - removing it wouldn't fix the behaviour.

### Minimal code example

```cpp
#include <fkYAML/node.hpp>
#include <string>

int main() {
    std::string input = "- !!map\n  key: value\n";
    fkyaml::node::deserialize(input);   // aborts in a Debug build
}
```

### Error messages

```Shell

```

MSVC; Windows 11

### Library version

0.4.4

### Validation

sndth:
Fixes #576.

A pending `!!map` tag survives until the next scalar token and is handed to `scalar_parser`, which
asserts that it never receives a collection tag. A collection tag on a scalar is a syntax error in the
input, not an internal inconsistency, so the deserializer now rejects it with a `parse_error` before
it gets that far. The assertions in `scalar_parser` are left in place: they are no longer reachable
from the deserializer, which is where an assertion belongs.

Removing the assertion instead would not have worked. With `NDEBUG` the parse continues and fails with
`Detected invalid indentation`, so it's a correct tripwire rather than an over-strict one.

Both scalar dispatch sites resolved the pending tag with the same four lines, so they now share a
`resolve_scalar_tag()` helper that carries the check.

### Scope

This stops the crash, it does not close the underlying gap. Of the four affected YAML test suite
cases, only `H7J7` is marked as an error case and it now passes. `6JWB`, `735Y` and `BU8L` are valid
YAML that should parse; they still fail, but with a `parse_error` rather than taking the process down.
Happy to open a follow-up for that separately.

### Verified

YAML test suite on `develop` vs. this branch: aborting cases 4 to 0, failing cases 113 down to 112,
with no newly failing case. Unit tests pass.

---

## Pull Request Checklist

## Please don't

### DIFF
--- a/tests/unit_test/test_deserializer_class.cpp
+++ b/tests/unit_test/test_deserializer_class.cpp
@@ -3491,6 +3491,19 @@ TEST_CASE("Deserializer_Tag") {
         auto input = GENERATE(std::string("foo: !!map !!map\n  bar: baz"), std::string("!!str !!bool true: 123"));
         REQUIRE_THROWS_AS(deserializer.deserialize(fkyaml::detail::input_adapter(input)), fkyaml::parse_error);
     }
+
+    SUBCASE("collection tag applied to a scalar node") {
+        // Only inputs where the tag really does apply to a scalar belong here. A `!!map` which precedes
+        // a block mapping, as in the `735Y` and `BU8L` suite cases, is valid YAML and must not be added.
+        auto input = GENERATE(
+            std::string("foo: !!seq bar"),
+            std::string("foo: !!seq \"bar\""),
+            std::string("foo: !!map 'bar'"),
+            std::string("foo: !!map |\n  bar"),
+            std::string("foo: !!seq >\n  bar"),
+            std::string("!!map foo"));
+        REQUIRE_THROWS_AS(deserializer.deserialize(fkyaml::detail::input_adapter(input)), fkyaml::parse_error);
+    }
 }
 
 TEST_CASE("Deserializer_NodeProperties") {
--- a/include/fkYAML/detail/input/deserializer.hpp
+++ b/include/fkYAML/detail/input/deserializer.hpp
@@ -1114,10 +1114,7 @@ private:
             case lexical_token_t::PLAIN_SCALAR:
             case lexical_token_t::SINGLE_QUOTED_SCALAR:
             case lexical_token_t::DOUBLE_QUOTED_SCALAR: {
-                tag_t tag_type {tag_t::NONE};
-                if (m_needs_tag_impl) {
-                    tag_type = tag_resolver_type::resolve_tag(m_tag_name, mp_meta);
-                }
+                const tag_t tag_type = resolve_scalar_tag(line, indent);
 
                 basic_node_type node = scalar_parser_type(line, indent).parse_flow(token.type, tag_type, token.str);
                 apply_directive_set(node);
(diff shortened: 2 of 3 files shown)


sndth:
Heads up: this and #575 are both branched off `09c5e36` and both add a private helper at the same
spot, just above `apply_node_properties`. Whichever you merge second will hit a trivial conflict
there, the two helpers are independent, they just land on the same line. Merge them in whatever
order suits you and I'll rebase the other one right away.

tests/unit_test/test_deserializer_class.cpp
```diff
@@ -3450,6 +3450,15 @@ TEST_CASE("Deserializer_Tag") {
         auto input = GENERATE(std::string("foo: !!map !!map\n  bar: baz"), std::string("!!str !!bool true: 123"));
         REQUIRE_THROWS_AS(deserializer.deserialize(fkyaml::detail::input_adapter(input)), fkyaml::parse_error);
     }
+
+    SUBCASE("collection tag applied to a scalar node") {
+        auto input = GENERATE(
+            std::string("foo: !!seq bar"),
+            std::string("foo: !!map |\n  bar"),
+            std::string("- !!map\n  key: value"),
+            std::string("key: &anchor\n !!map\n  foo: bar"));
```
fktn-k: Both `!!map`s are valid but applied to a mapping, not a scalar mapping key.  
`735Y` and `BU9L` in the YAML test suite use the same patterns and `test.event` shows they are associated to a mapping.  
So, it would be better to replace them with other cases or leave some comment that those `!!map` should be associated with a mapping according to the specification.

fktn-k (changes requested):
I've requested a change to the test cases.  
Please resolve the conflict as well.

tests/unit_test/test_deserializer_class.cpp
sndth: Oh, so sorry! I must have overlooked that - it even contradicts my own description above, where I
wrote those two are valid YAML which should parse.

Replaced them with inputs where the tag really does apply to a scalar, and left a comment so they do
not get added back. Conflict resolved too.

Please let me know if everything is fine now.

fktn-k: @sndth The changes are perfect. The updated test inputs now cover all the possible scalar styles.

fktn-k (approved):
LGTM

PR #579, opened 2026-08-22
Avoid passing a negative char to cctype functions

### LINKED ISSUE #578: A non-ASCII character in a tag URI is passed to std::isalnum as a negative char
### Description

Three call sites pass a plain `char` to a `cctype` function. When the character isn't representable
as `unsigned char` - any byte of a multi-byte UTF-8 sequence on a platform where `char` is signed - the
argument is negative and the call is undefined behaviour. MSVC's debug CRT asserts on it, glibc indexes
its table out of bounds.

| file | call |
|---|---|
| `include/fkYAML/detail/encodings/uri_encoding.hpp:123` | `std::isalnum(c)` |
| `include/fkYAML/detail/encodings/uri_encoding.hpp:69` | `std::tolower(*begin)` |
| `include/fkYAML/detail/input/lexical_analyzer.hpp:596` | `isalnum(*m_cur_itr)` |

`scalar_scanner.hpp` already avoids this with its own `is_digit()` and `is_xdigit()`, whose comments say
they exist "to avoid assertion failures in `std::isdigit()` especially when compiled with MSVC". These
three were missed.

Two further reasons to prefer explicit range checks over a cast here: the `lexical_analyzer.hpp` call is
unqualified, so it relies on the C header putting `isalnum` in the global namespace, and `std::isalnum`
is locale-dependent while a URI allows exactly `[0-9A-Za-z]` per RFC 3986.

I found by running the fuzz target with Asan, which reported this 50 times over a short session. Only the
`uri_encoding.hpp:123` site is demonstrated above; I couldn't construct an input reaching the other two,
but they are the same pattern and worth fixing together.

### Reproduction steps

Deserialize a tag whose URI contains a non-ASCII character, in a Debug build with MSVC.

Expected: the input is rejected with a `parse_error` and no diagnostic from the CRT.

Actual: the parse result is correct, but the debug CRT reports `Assertion failed: c >= -1 && c <= 255`
twice before the error is raised.

### Minimal code example

```cpp
#include <fkYAML/node.hpp>
#include <string>

int main() {
    std::string input = "!<tag:\xC3\xA9> foo";   // valid UTF-8 for "é"
    try { fkyaml::node::deserialize(input); } catch (const fkyaml::exception&) {}
}
```

### Error messages

```Shell

```

MSVC; Windows 11

### Library version

0.4.4

### Validation

sndth:
Fixes #578.

The three `cctype` calls are replaced with explicit ASCII range checks. Rather than spelling those out
at each site, they move into a new `detail/char_class.hpp` with `is_digit()`, `is_xdigit()` and
`is_alnum()`, and `scalar_scanner` now delegates to it instead of keeping its own two copies.

The shared header is the point rather than a tidy-up. The reason not to use `cctype` here had to be
written three times in my first attempt, and the two existing helpers were already a copy of each other,
which is how the bug arose in the first place: reaching for `std::isalnum` was the shortest path
because nothing local said otherwise. Now there is one place that says it.

As a side effect `validate_octets()` loses its `std::tolower` normalisation and the two range branches
that followed it, since `is_xdigit()` covers `A-F` directly.

### Verified

Behaviour is unchanged: the YAML test suite reports the same 115 failing cases as `develop`, with none
fixed and none newly failing. That is the intended result - this is an undefined behaviour fix, not
conformance one, so any movement there would mean I had altered parsing by accident.
Unit tests pass and the CRT reports nothing on the inputs that previously triggered two assertions each.

`test_uri_encoding_class.cpp` already covered `%AF%af`, so the octet change is guarded for case folding.
It had no byte above `0x7F` at all, so six were added, and two named tag handles with high bytes were
added to `test_lexical_analyzer_class.cpp`, which had no coverage of that validation.

---

## Pull Request Checklist

## Please don't

### DIFF
--- a/include/fkYAML/detail/char_class.hpp
+++ b/include/fkYAML/detail/char_class.hpp
@@ -1,0 +1,45 @@
+//  _______   __ __   __  _____   __  __  __
+// |   __| |_/  |  \_/  |/  _  \ /  \/  \|  |     fkYAML: A C++ header-only YAML library
+// |   __|  _  < \_   _/|  ___  |    _   |  |___  version 0.4.4
+// |__|  |_| \__|  |_|  |_|   |_|___||___|______| https://github.com/fktn-k/fkYAML
+//
+// SPDX-FileCopyrightText: 2023-2026 Kensuke Fukutani <fktn.dev@gmail.com>
+// SPDX-License-Identifier: MIT
+
+#ifndef FK_YAML_DETAIL_CHAR_CLASS_HPP
+#define FK_YAML_DETAIL_CHAR_CLASS_HPP
+
+#include <fkYAML/detail/macros/define_macros.hpp>
+
+FK_YAML_DETAIL_NAMESPACE_BEGIN
+
+// The functions below replace their <cctype> counterparts, which must not be used on the characters of
+// an input buffer. Those take an int whose value has to be representable as an unsigned char or equal
+// to EOF, so passing a char is undefined behavior for any byte of a multi-byte UTF-8 sequence on a
+// platform where char is signed. They also depend on the current locale, while both YAML syntax and
+// URI syntax allow exactly the ASCII ranges spelled out here.
+
+/// @brief Check if the given character is a digit.
+/// @param c A character to be checked.
+/// @return true if the given character is a digit, false otherwise.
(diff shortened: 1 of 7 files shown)


fktn-k (approved):
LGTM

PR #580, opened 2026-08-22
Fix the in-place benchmark parsing an empty buffer

sndth:
`bm_rapidyaml_parse_inplace` has been measuring nothing:

```cpp
std::string in_place_buff(test_src.size(), '\0');
c4::substr c4_test_src = c4::to_substr(in_place_buff).trimr('\0');
```

The view is taken before the loop, from a buffer that at that point consists entirely of `'\0'`, so
`trimr('\0')` strips all of it and leaves an empty view. The `memcpy` inside the loop fills the buffer,
but `c4_test_src` still spans zero bytes, so every iteration parses an empty input. Confirmed with a
temporary probe:

```
[PROBE] test_src=9119  c4_test_src=0
```

`SetBytesProcessed` credits each iteration with the full file size regardless, which is where the
implausible throughput comes from. On my machine, `ubuntu.yml` before and after:

| | before | after |
|---|---|---|
| `bm_rapidyaml_parse_inplace` | 216 ns, 40 GiB/s | 32153 ns, 293 MiB/s |
| `bm_rapidyaml_parse_arena` | 30567 ns, 283 MiB/s | 30731 ns, 282 MiB/s |

The two now land close together, which is what you would expect: parsing in place only saves one copy
of the source, which is small next to building the tree.

Initializing the buffer from `test_src` rather than from `'\0'` is enough - the `memcpy` in the loop
still restores it before each iteration, so the intent of the benchmark is unchanged.

The README claim that rapidyaml is "far faster with mutable buffers" rested entirely on this, so it is
removed. The rest of that sentence still holds: against `parse_in_arena` the ratio is about 2x in your
own numbers.

Two things I could not fix here, since they are measurements from your environment rather than mine:
the `rapidyaml (with mutable buff)` row in all three README tables, and the
`bm_rapidyaml_parse_inplace` lines in `tools/benchmark/results/`. Both would need regenerating.

---

## Pull Request Checklist

## Please don't

### DIFF
--- a/tools/benchmark/main.cpp
+++ b/tools/benchmark/main.cpp
@@ -103,8 +103,11 @@ void bm_rapidyaml_parse_inplace(benchmark::State& st) {
     for (auto _ : st) {
         // ryml::parse_in_place() modifies the contents of `in_place_buff` during parsing.
         // Without the following copy, the second (and subsequent) parsing would fail.
+        // The timer is paused around the copy so that only the parsing is measured.
+        st.PauseTiming();
         assert(in_place_buff.size() == test_src.size());
         std::memcpy(&in_place_buff[0], &test_src[0], in_place_buff.size());
+        st.ResumeTiming();
 
         ryml::Tree tree = ryml::parse_in_place(c4_test_src);
     }
--- a/tools/benchmark/main.cpp
+++ b/tools/benchmark/main.cpp
@@ -94,7 +94,10 @@ void bm_libfyaml_parse(benchmark::State& st) {
 
 // rapidyaml (in place)
 void bm_rapidyaml_parse_inplace(benchmark::State& st) {
-    std::string in_place_buff(test_src.size(), '\0');
+    // Initialize from the source so the view below spans the real contents. Constructing the buffer
+    // filled with '\0' instead would make trimr('\0') strip all of it, leaving an empty view and
+    // making the benchmark parse nothing.
+    std::string in_place_buff(test_src);
     c4::substr c4_test_src = c4::to_substr(in_place_buff).trimr('\0');
 
     for (auto _ : st) {
--- a/README.md
+++ b/README.md
@@ -224,31 +224,31 @@ The following tables are created from the benchmarking results in the following 
 
 | Benchmark                          | processed bytes per second (Release) |
 | ---------------------------------- | ------------------------------------ |
-| fkYAML                             | 62.5049Mi/s                          |
-| libfyaml                           | 39.235Mi/s                           |
-| rapidyaml<br>(with mutable buff)   | 22.007Gi/s                           |
-| rapidyaml<br>(with immutable buff) | 133.311Mi/s                          |
-| yaml-cpp                           | 9.07876Mi/s                          |
+| fkYAML                             | 60.6026Mi/s                          |
+| libfyaml                           | 35.4699Mi/s                          |
+| rapidyaml<br>(with mutable buff)   | 152.795Mi/s                          |
+| rapidyaml<br>(with immutable buff) | 136.538Mi/s                          |
+| yaml-cpp                           | 10.3192Mi/s                          |
 
 ### Parsing [citm_catalog.json](https://github.com/fktn-k/fkYAML/blob/develop/tools/benchmark/cases/citm_catalog.json)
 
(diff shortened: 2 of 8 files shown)


tools/benchmark/main.cpp
fktn-k, line 109: I realized that the `assert` and `std::memcpy` calls should be excluded from the benchmark to measure pure parsing performance.  
The benchmarking tool in the rapidyaml repository does that by wrapping those operations with `st.PauseTiming()` and `st.ResumeTiming()`.  
https://github.com/biojppm/rapidyaml/blob/master/bm/bm_common.hpp#L240

fktn-k (changes requested):
Although the reviewed point is irrelevant to your change, it would be nice if you could fix it too.

fktn-k:
@sndth  
> Two things I could not fix here, since they are measurements from your environment rather than mine:
the rapidyaml (with mutable buff) row in all three README tables, and the
bm_rapidyaml_parse_inplace lines in tools/benchmark/results/. Both would need regenerating.

I'm happy to take that part. After I approve the changes, I will make a commit to update them and share it with you. I would appreciate it if you `cherry-pick` it to your branch.

sndth:
> @sndth
> 
> > Two things I could not fix here, since they are measurements from your environment rather than mine:
> > the rapidyaml (with mutable buff) row in all three README tables, and the
> > bm_rapidyaml_parse_inplace lines in tools/benchmark/results/. Both would need regenerating.
> 
> I'm happy to take that part. After I approve the changes, I will make a commit to update them and share it with you. I would appreciate it if you `cherry-pick` it to your branch.

Good catch, thank you. The reset is now wrapped in `st.PauseTiming()` / `st.ResumeTiming()`,
the same way `BmCase::prepare()` does it in the rapidyaml benchmarks.

I measured it before and after, 5 repetitions each (MSVC/Windows here, so the absolute
numbers are not comparable to yours): the reset was worth about 520 ns per iteration on
`ubuntu.yml` (1.8%), and stays below the noise on `citm_catalog.yml`. `PauseTiming()`
itself did not cost anything measurable at these parse times.

Thank you also for offering to regenerate the rapidyaml numbers. I have left the README
rows and the files in `tools/benchmark/results/` untouched, so your commit will apply
cleanly for me to cherry-pick.

tools/benchmark/main.cpp
fktn-k: @sndth Thank you for the fix! Confirmed.  
https://github.com/fktn-k/fkYAML/pull/580#issuecomment-5388512744

fktn-k (approved):
LGTM

fktn-k:
@sndth  
I updated the benchmark results in the [update_benchmark_results](https://github.com/fktn-k/fkYAML/tree/update_benchmark_results) branch whose base branch is your fix_inplace_benchmark branch so `cherry-pick`ing the commit can be applied cleanly.

sndth:
> @sndth I updated the benchmark results in the [update_benchmark_results](https://github.com/fktn-k/fkYAML/tree/update_benchmark_results) branch whose base branch is your fix_inplace_benchmark branch so `cherry-pick`ing the commit can be applied cleanly.

Cherry-picked! Thank you for regenerating them. The rapidyaml rows now land in the same range as the
immutable-buffer ones, which is what I'd expect once the benchmark stops parsing an empty buffer.

I added one follow-up commit. The summary sentence under the tables no longer matched the new
numbers, so I recomputed all three claims against them. Each cell below is the ratio of the first
library's throughput to the second's:

| | rapidyaml / fkYAML | fkYAML / libfyaml | fkYAML / yaml-cpp |
|---|---|---|---|
| `ubuntu.yml` | 2.25-2.52x | 1.71x | 5.87x |
| `citm_catalog.json` | 1.22-1.24x | 1.93x | 5.81x |
| `citm_catalog.yml` | 1.30-1.31x | 1.86x | 5.52x |

"about 2x faster" only holds for `ubuntu.yml`, "70% faster than libfyaml" understates it, and
"about 6.5x faster than yaml-cpp" overstates it. The sentence now reads "1.2x to 2.5x faster",
"70-90% faster" and "more than 5x faster". Happy to reword if you'd rather phrase it differently.

fktn-k:
@sndth  
The followup commit is a great help. No rewording is necessary. Thanks you so much!

PR #581, opened 2026-08-22
Allocate node properties and document metainfo only when needed

sndth:
This started as a question rather than a plan. While reading through the deserializer I noticed that
`node_property` - the anchor and the tag, so two `std::string`s - sits inline in every `basic_node`,
whether or not anything is ever set on it. Almost no node in a real document carries either. Same for
`document_metainfo`, which was being created per node instead of once.

So I wondered what would happen if those were separated out and allocated only when something actually
needs them. Both now live behind a pointer created on first use.

I expected a few percent:

| | before | after | |
|---|---|---|---|
| `ubuntu.yml` | 46.9 MiB/s | **86.5 MiB/s** | 1.8x |
| `citm_catalog.yml` | 26.2 MiB/s | **55.5 MiB/s** | 2.1x |

I did not expect that 😮

10 repetitions per side, same binary and same machine, differing only in the fkYAML headers; the
medians agree with the means. This is MSVC on Windows, so the absolute figures will not match your
Ubuntu results, but the effect is a matter of allocation counts rather than of the compiler.

### Where the factor comes from

The speedup is not a cleverer algorithm - it is the parser no longer allocating. Counting every
`operator new` across a single parse:

| | allocations | peak heap |
|---|---|---|
| `ubuntu.yml` | 3 897 -> **1 068** (-73%) | 98 KiB -> **61 KiB** (-37%) |
| `citm_catalog.yml` | 514 061 -> **108 364** (-79%) | 11.04 MiB -> **6.39 MiB** (-42%) |

These counts are deterministic, so they need no repetitions. The absolute numbers depend on the
standard library's small-string threshold; the ratio does not.

`sizeof(fkyaml::node)` goes from 96 to 40 bytes, and a node with no anchor and no tag no longer
constructs and destroys two strings. On `citm_catalog.yml` that multiplies across hundreds of
thousands of nodes, which is where the factor comes from.

### Behaviour is unchanged

A parser which is faster because it quietly does less work would be an easy mistake to make here, so
that is the check I care about most: YAML test suite reports the same 113 failing cases as
`develop`, identical case for case. Unit tests also pass.

Anchor and tag semantics are unaffected. The copy constructor deep-copies the property, so value
semantics are preserved, and neither `node_property` nor `document_metainfo` appears in a public
signature. Since the metainfo is now created on first use, a node built outside a parsed document has
none, so tag resolution falls back to the default handle prefixes rather than assuming a metainfo
exists.

A node which does carry an anchor or a tag now pays one extra allocation and one indirection - the
rare case gets slightly slower so that the common one gets much faster.

---

## Pull Request Checklist

## Please don't

### DIFF
--- a/README.md
+++ b/README.md
@@ -224,33 +224,33 @@ The following tables are created from the benchmarking results in the following 
 
 | Benchmark                          | processed bytes per second (Release) |
 | ---------------------------------- | ------------------------------------ |
-| fkYAML                             | 60.6026Mi/s                          |
-| libfyaml                           | 35.4699Mi/s                          |
-| rapidyaml<br>(with mutable buff)   | 152.795Mi/s                          |
-| rapidyaml<br>(with immutable buff) | 136.538Mi/s                          |
-| yaml-cpp                           | 10.3192Mi/s                          |
+| fkYAML                             | 77.8628Mi/s                          |
+| libfyaml                           | 34.7381Mi/s                          |
+| rapidyaml<br>(with mutable buff)   | 150.988Mi/s                          |
+| rapidyaml<br>(with immutable buff) | 135.589Mi/s                          |
+| yaml-cpp                           | 10.1239Mi/s                          |
 
 ### Parsing [citm_catalog.json](https://github.com/fktn-k/fkYAML/blob/develop/tools/benchmark/cases/citm_catalog.json)
 
 | Benchmark                          | processed bytes per second (Release) |
 | ---------------------------------- | ------------------------------------ |
-| fkYAML                             | 97.4616Mi/s                          |
-| libfyaml                           | 50.3768Mi/s                          |
-| rapidyaml<br>(with mutable buff)   | 120.381Mi/s                          |
-| rapidyaml<br>(with immutable buff) | 119.205Mi/s                          |
-| yaml-cpp                           | 16.7686Mi/s                          |
+| fkYAML                             | 126.783Mi/s                          |
+| libfyaml                           | 49.9466Mi/s                          |
+| rapidyaml<br>(with mutable buff)   | 120.681Mi/s                          |
+| rapidyaml<br>(with immutable buff) | 118.728Mi/s                          |
+| yaml-cpp                           | 16.5976Mi/s                          |
 
 ### Parsing [citm_catalog.yml](https://github.com/fktn-k/fkYAML/blob/develop/tools/benchmark/cases/citm_catalog.yml)
 
 | Benchmark                          | processed bytes per second (Release) |
 | ---------------------------------- | ------------------------------------ |
-| fkYAML                             | 41.4116Mi/s                          |
-| libfyaml                           | 22.3071Mi/s                          |
-| rapidyaml<br>(with mutable buff)   | 54.1632Mi/s                          |
-| rapidyaml<br>(with immutable buff) | 53.6708Mi/s                          |
-| yaml-cpp                           | 7.4989Mi/s                           |
+| fkYAML                             | 55.9384Mi/s                          |
+| libfyaml                           | 21.6074Mi/s                          |
+| rapidyaml<br>(with mutable buff)   | 54.2151Mi/s                          |
+| rapidyaml<br>(with immutable buff) | 53.7195Mi/s                          |
+| yaml-cpp                           | 7.0331Mi/s                           |
 
-Although [rapidyaml](https://github.com/biojppm/rapidyaml) is 1.2x to 2.5x faster than fkYAML as it focuses on high performance, fkYAML is in general 70-90% faster than [libfyaml](https://github.com/pantoniou/libfyaml) and also more than 5x faster than [yaml-cpp](https://github.com/jbeder/yaml-cpp).  
+Although [rapidyaml](https://github.com/biojppm/rapidyaml) focuses on high performance and can be up to about 1.9x faster than fkYAML depending on the input, fkYAML is in general 2.2x to 2.6x faster than [libfyaml](https://github.com/pantoniou/libfyaml) and also more than 7x faster than [yaml-cpp](https://github.com/jbeder/yaml-cpp).  
 Note that, since fkYAML deserializes scalars into native booleans or integers during the parsing, the performance could be more faster in some use cases since there is no need for string manipulations upon data queries.  
 
 ## Community Support
--- a/docs/docs/api/basic_node/set_yaml_version.md
+++ b/docs/docs/api/basic_node/set_yaml_version.md
@@ -3,7 +3,7 @@
 # <small>fkyaml::basic_node::</small>set_yaml_version
 
 ```cpp
-void set_yaml_version(const yaml_version_t version) noexcept;
+void set_yaml_version(const yaml_version_t version);
 ```
 
 !!! warning "Deprecation"
--- a/docs/docs/api/basic_node/set_yaml_version_type.md
+++ b/docs/docs/api/basic_node/set_yaml_version_type.md
@@ -3,7 +3,7 @@
 # <small>fkyaml::basic_node::</small>set_yaml_version_type
 
 ```cpp
-void set_yaml_version_type(const yaml_version_type version) noexcept;
+void set_yaml_version_type(const yaml_version_type version);
 ```
 
 Sets a target YAML specification version to the `basic_node` object.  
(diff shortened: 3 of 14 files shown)


sndth:
Codacy reports `Complexity increased by 21 (20 max.)` on `node.hpp`, and I could use your call on it.

I tried to bring it down. Collapsing three null guards in `get_resolved_tag_name()` into a single
fallback lowered that function's cyclomatic complexity from 12 to 10, and the file's added decision
points from 9 to 7. Codacy still reports exactly 21, unchanged. So the number does not seem to track
the branches this PR adds, and I have no reliable way to predict what would move it.

The increase is inherent to the change either way. `document_metainfo` and `node_property` used to be
unconditionally present, so every read could dereference them directly. Creating them on first use
removes that guarantee, so each read gains a null check - 10 across `node.hpp` and `serializer.hpp`.
What they buy, measured on the benchmark cases:

| | allocations | peak heap | parse |
|---|---|---|---|
| `ubuntu.yml` | 3 897 -> **1 068** | 98 -> **61 KiB** | **1.9x** |
| `citm_catalog.yml` | 514 061 -> **108 364** | 11.04 -> **6.39 MiB** | **2.0x** |

I can split this into two PRs - the commits are already independent, one for the metainfo and one for
the node properties, and each builds and passes on its own. That would shrink each diff, though I
cannot promise it clears the gate either.

Every recent PR here has been green on Codacy, so I would rather ask than assume: is it blocking for
you, or advisory? If you want it green, I will do the split.

sndth:
I haven't finished all the changes in this PR yet. I'm waiting for all the workflows to complete and I'll let you know later but I'm open to review at this stage.

sndth:
Following up on my note above: the changes are complete now!

Three things landed since then:

- The `get_resolved_tag_name()` guard moved into the first commit, where the nullable metainfo is
  introduced. Previously the first commit dereferenced `mp_meta` unconditionally while already
  allowing it to be null, so that commit alone crashed. No commit on the branch does now.
- `set_yaml_version_type()` keeps its `noexcept`, so the public API is unchanged, I diffed every
  public signature against `develop` and there is no difference. More on this below.
- Two tests: tag resolution on a node that was never part of a parsed document, and a copy-constructor
  test asserting the tag is deep-copied rather than shared.

The branch is rebased onto current `develop` also.

### Verification

Beyond the unit tests and the YAML test suite, I ran a differential check: the same program compiled
once against `develop`'s headers and once against this branch's, over all 402 `in.yaml` files from
the test suite, comparing the parse result, the serialization, the YAML version, the resolved tag and
the anchor for every document.

| | result |
|---|---|
| differential, 402 suite documents | byte-identical |
| ASan, same 402 documents + API exercise | no reports |
| leak check, both branches | heap returns to zero after every document |
| libFuzzer + ASan, seeded with the suite | 837 918 executions in 181 s, no crashes |

The leak check is separate because ASan on Windows has no LeakSanitizer; I counted `operator new` /
`delete` instead. libFuzzer grew the corpus from 402 to 2 908 inputs and produced no artifacts.

### A question on `set_yaml_version_type()`

This is the one place where behaviour is not strictly identical to `develop`, and I would rather you
decide than guess.

On `develop` the function was `noexcept` and genuinely could not fail - the metainfo always existed,
so it only assigned two fields. Now the metainfo is created on first use, so the function allocates.
I kept `noexcept`, because removing it from a public function in a minor release seemed worse than
the alternative. The consequence is that on allocation failure `std::bad_alloc` escapes a `noexcept`
function and calls `std::terminate()` instead of propagating. The deprecated `set_yaml_version()`
forwards to it and has the same property.

So the options as I see them:

1. Keep `noexcept` (current): API identical to `develop`; OOM terminates.
2. Drop `noexcept` from both: `bad_alloc` propagates; the signature changes, which may break code
   that relies on the contract.
3. Something else: if you would rather the version not live behind the lazily created metainfo at
   all, I am happy to look into it.

Which would you prefer?

fktn-k:
@sndth  
Thank you for this great work!  
I was truly amazed at how dramatically the performance improved by reducing the number of allocations and the node size :tada:  

This is just my guess but it's likely that unnecessary allocations, indirect references, or cache misses were causing the performance to slow down...? :thinking:  

---

These are answers to your questions.  

**Codacy report**:  
It is an advisory for me, not a requirement, since a large patch is more likely to cause the increase in complexity than a small one. As long as I see the changes, they are all reasonable and not so complex that I request changes.  

**set_yaml_version_type()**:  
I would prefer the second option, Drop `noexcept` from both.  
For the first option, leaving `noexcept` in place causes `std::terminate` to be called when an exception occurs, which is akin to the crashes caused by assertion failures—which we recently fixed for parsing—still occurring in the release version. This means that performance improvements would come at the cost of making the software unstable.  
For the third option, I prefer keeping the version in `document_metainfo` since it is a document-level information like `%TAG` directives. Actually, I'm thinking of changes in the current data structure to separate document-level information from node-level information, like a `document` class instance manages its associated `node` instances since it is very strange that document-level information is owned by `node` instances, which should be the opposite way.

sndth:
> @sndth Thank you for this great work! I was truly amazed at how dramatically the performance improved by reducing the number of allocations and the node size 🎉
> 
> This is just my guess but it's likely that unnecessary allocations, indirect references, or cache misses were causing the performance to slow down...? 🤔
> 
> These are answers to your questions.
> 
> **Codacy report**: It is an advisory for me, not a requirement, since a large patch is more likely to cause the increase in complexity than a small one. As long as I see the changes, they are all reasonable and not so complex that I request changes.
> 
> **set_yaml_version_type()**: I would prefer the second option, Drop `noexcept` from both. For the first option, leaving `noexcept` in place causes `std::terminate` to be called when an exception occurs, which is akin to the crashes caused by assertion failures—which we recently fixed for parsing—still occurring in the release version. This means that performance improvements would come at the cost of making the software unstable. For the third option, I prefer keeping the version in `document_metainfo` since it is a document-level information like `%TAG` directives. Actually, I'm thinking of changes in the current data structure to separate document-level information from node-level information, like a `document` class instance manages its associated `node` instances since it is very strange that document-level information is owned by `node` instances, which should be the opposite way.

Thank you for the review, and for the detailed reasoning on `noexcept` - that settles it.

`noexcept` is dropped from both `set_yaml_version_type()` and the deprecated `set_yaml_version()`,
so `std::bad_alloc` now propagates instead of reaching `std::terminate()`. The getters keep it,
since they only read the metainfo and never create it.

I didn't add tests for the throwing path. The change removes an attribute rather than a code path,
and the only exception that can escape is `std::bad_alloc` from the `make_shared` call in `meta()`.
Since `basic_node` has no allocator parameter, that allocation goes through the global
`operator new`, so there is no injection point short of overriding it for the whole test binary.
Happy to add something if you would like it covered anyway.

On the wider change you mentioned - separating document-level information from node-level
information - I agree the current ownership is the wrong way around, and I would be glad to help
with it whenever you decide to start.

sndth:
One note on the numbers above: they were measured on Windows with MSVC. Part of the gain comes from
avoiding allocator overhead, which is cheaper on glibc, so the ratio may be lower on your setup.

fktn-k (approved):
@sndth  
The changes look good to me.  

As for testing `std::bad_alloc` possibly thrown from `set_yaml_version_type` and `set_yaml_version`, I'm fine with letting it as-is in the unit tests. One way I can come up with is that overriding the global `void* operator new(std::size_t size)` where a singleton flag is tested to decide whether it should throw an `std::bad_alloc` no matter what the runtime environment is. However, I think it goes too far given the methods just propagate the exception to the caller.

fktn-k:
I'll hold this PR. I would like to update the benchmark results in here again to reflect the current status in the documentation.  
So, I'd like to ask you to rebase this branch onto the develop HEAD once #580 is merged, and `cherry-pick` a commit where I'll regenerate the results after rebasing.

sndth:
Sounds good. I'll rebase onto `develop` as soon as #580 is merged and let you know here, so you
can regenerate the results on top of it.

fktn-k:
@sndth  
Thank you for taking this on. I just merged #580.  Please let me know when you finish rebasing.

sndth:
> @sndth Thank you for taking this on. I just merged #580. Please let me know when you finish rebasing.

I just rebased onto `develop` at 6d42e43, now that #580 is merged.
Ready for your commit with the regenerated results whenever it suits you I'll cherry-pick it.

fktn-k:
@sndth 
I finished updating the benchmark results and committed them here:  
https://github.com/fktn-k/fkYAML/tree/update_benchmark_results_after_refactoring  

While it depends on the input, I'm pleased to see that performance has improved dramatically thanks to your patches.

sndth:
Cherry-picked, thanks! Those numbers are better than I dared hope.

fktn-k:
Thanks!

PR #582, opened 2026-08-23
Add a workflow job to run the unit tests with MSVC sanitizers

sndth:
This is a follow-up to @fktn-k's suggestion in a comment on #578:

> Now we use sanitizers for the fuzzing test, it would be nice if we have a workflow to run unit tests
> using MSVC with sanitizers enabled.

This adds a `FK_YAML_RUN_MSVC_SANITIZERS` option and an `msvc_sanitizers` job in `windows.yml` which
builds and runs the unit tests under MSVC's AddressSanitizer. The compile and link flags are the ones
the fuzz test already uses on MSVC, added in #571.

One part is not obvious from the diff: `ctest` runs the test executable outside the Visual Studio
environment, where `clang_rt.asan_dynamic-x86_64.dll` is not on `PATH`, so the executable fails to
start before a single test runs. The runtime ships next to `cl.exe`, so `tests/unit_test/CMakeLists.txt`
copies it beside the executable after the build - the same thing `tests/fuzz_test/CMakeLists.txt`
already does. Keeping it in CMake rather than the workflow means the local instructions in
`CONTRIBUTING.md` work too.

Your other note in #571, about a dedicated workflow for the fuzz tests, is a separate change - I will
open it separately.

---

## Pull Request Checklist

## Please don't


fktn-k (approved):
LGTM. Thank you for your prompt response!

PR #587, opened 2026-08-23
Support a line break between a flow mapping key and its value indicator

sndth:
In the flow context, YAML allows the `:` indicator to be separated from its key by a line break -
`ns-flow-map-yaml-key-entry` permits `s-separate` before `c-ns-flow-map-separate-value`. The
deserializer required both to be on the same line, which is only correct for implicit keys in the
block context:

```yaml
{"foo"
: "bar"}
```

Relaxing that on its own would accept under-indented flow contents, so this also adds the rule that
comes with it: the contents of a flow context nested in a block context must be more indented than
that block context, while a flow context beginning at the document level stays unconstrained. Without
it, `VJP3/00` from the test suite, which is an error case, would no longer be rejected.

## Effect on yaml-test-suite

Against `data-2022-01-17`, failing cases go from 110 to 103, with no regressions:

| case | |
|---|---|
| `4MUZ/00`, `4MUZ/01`, `4MUZ/02` | flow mapping colon on the line after the key |
| `NJ66`, `9SA2` | multiline flow mapping key |
| `VJP3/01` | flow collections over many lines |
| `9C9N` | wrong indented flow sequence, now rejected |

Three cases still fail with the same error message for unrelated reasons, and aren't addressed here:
`5MUD` and `K3WX` need `:` adjacent to a value after a JSON-like key, and `UT92` needs `%` to be
treated as plain content inside a flow context.

## What with tests?

`Deserializer_FlowMapping` gains cases for the separated indicator and for a key spanning two lines.
A new `Deserializer_FlowContentIndentation` covers under-indented contents, correctly indented
contents, and a document-level flow context whose contents start at column 0.

---

## Pull Request Checklist

## Please don't

### DIFF
--- a/tests/unit_test/test_deserializer_class.cpp
+++ b/tests/unit_test/test_deserializer_class.cpp
@@ -2556,6 +2556,27 @@ TEST_CASE("Deserializer_FlowMapping") {
         REQUIRE(foo_node.is_mapping());
         REQUIRE(foo_node.empty());
     }
+
+    SUBCASE("key separator on the line after the key") {
+        auto input =
+            GENERATE(std::string("{\"foo\"\n: bar}"), std::string("{foo\n: bar}"), std::string("{foo\n:\nbar}"));
+        REQUIRE_NOTHROW(root = deserializer.deserialize(fkyaml::detail::input_adapter(input)));
+
+        REQUIRE(root.is_mapping());
+        REQUIRE(root.size() == 1);
+        REQUIRE(root.contains("foo"));
+        REQUIRE(root["foo"].as_str() == "bar");
+    }
+
+    SUBCASE("quoted key spanning multiple lines") {
+        auto input = GENERATE(std::string("{ \"multi\n  line\": value }"), std::string("{ 'multi\n  line': value }"));
+        REQUIRE_NOTHROW(root = deserializer.deserialize(fkyaml::detail::input_adapter(input)));
+
+        REQUIRE(root.is_mapping());
+        REQUIRE(root.size() == 1);
+        REQUIRE(root.contains("multi line"));
+        REQUIRE(root["multi line"].as_str() == "value");
+    }
 }
 
 TEST_CASE("Deserializer_UnclosedFlowCollection") {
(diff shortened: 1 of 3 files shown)


sndth:
@fktn-k

Unrelated to this PR, but I ran into it while going through the test suite: `deserialize_docs()`
returns a single null document for a stream that contains only comments, whitespace or a bare `...`,
where the suite expects none (`AVM7`, `HWV9`, `QT73`, `8G76`, `98YD`), and returns one document for
`---\n---` where it expects two (`6XDY`). Fixing it would change what `deserialize_docs()` returns
for those inputs, so I'd rather ask than assume - is that a change you would want?

tests/unit_test/test_deserializer_class.cpp
```diff
@@ -2556,6 +2556,27 @@ TEST_CASE("Deserializer_FlowMapping") {
...
+        REQUIRE_NOTHROW(root = deserializer.deserialize(fkyaml::detail::input_adapter(input)));
+
+        REQUIRE(root.is_mapping());
+        REQUIRE(root.size() == 1);
+        REQUIRE(root.contains("foo"));
+        REQUIRE(root["foo"].as_str() == "bar");
+    }
+
+    SUBCASE("key spanning multiple lines") {
+        std::string input = "{ multi\n  line: value }";
```
fktn-k: Although the current implementation allows this, such a plain scalar key is actually invalid according to the YAML specification. Could you wrap the key with either single or double quotes instead? A quoted key can span multiple lines.

fktn-k (changes requested):
I've requested a small change to the added test case.

fktn-k:
> @fktn-k
> 
> Unrelated to this PR, but I ran into it while going through the test suite: `deserialize_docs()` returns a single null document for a stream that contains only comments, whitespace or a bare `...`, where the suite expects none (`AVM7`, `HWV9`, `QT73`, `8G76`, `98YD`), and returns one document for `---\n---` where it expects two (`6XDY`). Fixing it would change what `deserialize_docs()` returns for those inputs, so I'd rather ask than assume - is that a change you would want?

@sndth  
Regarding empty documents, the current node implementation has no way to express an empty document. So, I think it's ok for now to regard them as allowed failures until some document-level information is extracted as a separate class.

tests/unit_test/test_deserializer_class.cpp
sndth: Changed! The key is now quoted, and the case covers both double and single quoting.

One thing worth flagging, since it follows from your point: the plain multi-line key is `NJ66` in
the test suite, which this PR makes pass and which the suite expects to succeed. `9SA2` is the same
construct with a quoted key, and the suite expects that to pass too, so the suite doesn't seem to
draw the distinction. If you read the single-line restriction as applying to plain keys, then fkYAML
now accepts something it should reject, and `NJ66` becomes an allowed failure rather than a fix.

I haven't changed the implementation either way, tell me which you would prefer.

fktn-k: The change looks good. Thanks!  


As for multi-line plain scalar keys, both lexical_analyzer and scalar_parser have no knowledge of whether their pricessing scalar is a key. So, although it seems correctly throwing an exception, I assume it would not be an expected result. I would like to know what the current implementation treats as an error in `NJ66` and correct it, of course in a separate task.

fktn-k (approved):
LGTM

tests/unit_test/test_deserializer_class.cpp
sndth: Thanks for merging!

On `NJ66`: I checked what the current implementation does on `develop` after this PR, and the answer
is that it treats nothing as an error there. It parses cleanly and produces `multi line` as the key:

```
NJ66 (plain multi-line key, flow):    OK  -> "multi line": value
9SA2 (quoted multi-line key, flow):   OK  -> "multi line": value
```

So it is not a matter of correcting an exception - it currently accepts the construct. The same key
in a block context does throw, but with a misleading message:

```
multi
  line: value
-> parse_error: invalid explicit mapping key separator is found. (at line 1, column 6)
```

On the obstacle you mention: `lexical_analyzer` and `scalar_parser` indeed have no idea whether a
scalar is a key, but the deserializer does, and it still has the token type at that point. The
`PLAIN_SCALAR` / `SINGLE_QUOTED_SCALAR` / `DOUBLE_QUOTED_SCALAR` case in `deserialize_scalar()` is
the same place that decides the scalar becomes a key via `add_new_key()`, so a "this plain scalar
spans lines" flag computed there would have everything it needs.

One thing worth settling before that task: §6.7 says "Implicit keys are restricted to a single line"
without distinguishing plain from quoted, yet the suite expects both `NJ66` and `9SA2` to pass. If
the restriction is applied as written, `9SA2` would have to be rejected too.

fktn-k: I'm not sure but the context (block or flow) seems matter. (Is it ok to use an implicit multiline scalar key in flow contexts but not in block context..?)

sndth:
> > @fktn-k
> > Unrelated to this PR, but I ran into it while going through the test suite: `deserialize_docs()` returns a single null document for a stream that contains only comments, whitespace or a bare `...`, where the suite expects none (`AVM7`, `HWV9`, `QT73`, `8G76`, `98YD`), and returns one document for `---\n---` where it expects two (`6XDY`). Fixing it would change what `deserialize_docs()` returns for those inputs, so I'd rather ask than assume - is that a change you would want?
> 
> @sndth Regarding empty documents, the current node implementation has no way to express an empty document. So, I think it's ok for now to regard them as allowed failures until some document-level information is extracted as a separate class.

@fktn-k 
The suite is unambiguous on this, and the line runs slightly differently from block vs flow, it's
"flow mapping entry" vs "implicit key":

| case | input | expected |
|---|---|---|
| `NJ66` | `- { multi` / `  line: value}` | valid |
| `9SA2` | `- { "multi` / `  line": value}` | valid |
| `C2SP` | `[23` / `]: 42` | error |
| `DK4H` | `[ key` / `  : value ]` | error |

The key of a flow mapping entry is a full flow node, so it may span lines, which is why both `NJ66`
and `9SA2` are expected to parse, plain and quoted alike. The single-line restriction applies to
implicit keys, and those appear in block mappings (`C2SP`) and in single pair mappings inside flow
sequences (`DK4H`), but not in `{...}` entries.

So on the accept/reject axis `develop` is already right for `NJ66` and `9SA2`. What is off is only
the diagnostic in the block case, which reports something unrelated:

```
multi
  line: value
-> parse_error: invalid explicit mapping key separator is found. (at line 1, column 6)
```

`DK4H` is the one that doesn't match yet: `develop` accepts `[ key\n  : value ]`, where the
restriction says it should be rejected.

PR #589, opened 2026-08-24
Fix the indentation base of a document level block scalar

sndth:
A block scalar which is the root node of a document has no parent, so per the spec its contents are
measured against `n = -1` and may begin at the first column:

```yaml
--- >
line1
line2
```

The lexer measured them against the indentation of the header line, which is `0` for both `--- >`
and a bare `>`, so `base_indent >= cur_indent` rejected any content starting at column 0.

A block scalar counts as a document root only when nothing but an optional `---` marker precedes its
header on the line, so a header sitting on an indented line of its own is still measured against its
parent.

## Tests

YAML test suite failing cases go from 102 to 100 with no regressions: `FP8R` and `DK3J` now pass. `M7A3` and `W4TN`
parse correctly now too, but still fail on the document count, which per @fktn-k's answer in #587
stays an allowed failure for now.

I also added `Deserializer_DocumentLevelBlockScalar`, which covers a root block scalar starting at the first column
(`--- >` and a bare `>`), an indented one, and the negative cases `foo: >` and `- |`.

---

## Pull Request Checklist

## Please don't

### DIFF
--- a/include/fkYAML/detail/input/lexical_analyzer.hpp
+++ b/include/fkYAML/detail/input/lexical_analyzer.hpp
@@ -382,6 +382,8 @@ private:
             }
 
             const uint32_t base_indent = get_current_indent_level(&sv[header_end_pos]);
+            // Must be checked before m_token_begin_itr is moved to the beginning of the contents below.
+            const bool is_document_root = begins_document_level_node();
 
             const str_view header_line = sv.substr(1, header_end_pos - 1);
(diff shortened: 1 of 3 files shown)


sndth:
A milestone worth a mention! This brings the yaml-test-suite failures to a round 100 - it was 110 a
few days ago! 🎉

fktn-k (approved):
LGTM. Thanks for the clean solution!

PR #590, opened 2026-08-25
Continue a multiline plain scalar across equally indented lines

sndth:
A plain scalar which begins a line of its own can be continued by lines at the same indentation,
since nothing else on that line owns it:

```yaml
Warning:
  This is an error message
  for the log file
```

The continuation was measured against the indentation of the scalar's own first line, so a line at
that same indentation ended it. The scalar then broke into one token per line, and the deserializer
reported a misleading `The ":" mapping value indicator must be followed after a mapping key.`

A scalar which follows a key on the same line still requires more indented lines, because a line at
the key's indentation belongs to the parent mapping instead. The two cases are told apart by
comparing the scalar's column with the indentation of its line.

Lowering the bar also required naming the things which end a plain scalar explicitly, since the old
rule had been stopping scalars at column 0 as a side effect: a document marker, a `:` or `?`
followed by a space, and a flow collection. `- ` is deliberately not among them, because an indented
one is legitimate content (`AB8U`).

I ran tests and the YAML test suite failing cases go from 100 to 95 with no regressions: `4CQQ`, `NB6Z`, `RZT7` and
`XLQ9` now parse, and `DK4H` is now correctly rejected.

I also added `Deserializer_MultilinePlainScalarInBlockContext`, covering a scalar continued at the same
indentation, one following its key which needs deeper indentation, a document marker ending the
scalar, and `foo\n: bar` still being rejected. One existing expectation in
`LexicalAnalyzer_PlainScalar` changed: `"  foo\n   bar\n  baz"` is now a single token, which is the
same shape the suite requires in `4CQQ`.

---

## Pull Request Checklist

## Please don't

### DIFF
--- a/include/fkYAML/detail/input/lexical_analyzer.hpp
+++ b/include/fkYAML/detail/input/lexical_analyzer.hpp
@@ -430,6 +430,48 @@ private:
         return true;
     }
 
+    /// @brief Checks if the given position is a white space, a line break or the end of the buffer.
+    /// @note Tabs count as white space here, as in the definition of c-forbidden.
+    /// https://yaml.org/spec/1.2.2/#912-document-markers
+    /// @param sv The buffer contents being scanned.
+    /// @param pos The position to inspect.
+    /// @return true if nothing, a white space or a line break follows, false otherwise.
+    static bool is_followed_by_white_space(str_view sv, std::size_t pos) noexcept {
+        return (pos == sv.size()) || (sv[pos] == ' ') || (sv[pos] == '\t') || (sv[pos] == '\n');
+    }
+
+    /// @brief Checks if the given position begins something a plain scalar cannot continue into.
+    /// @param sv The buffer contents being scanned.
+    /// @param pos The position of the first non-space character in a line.
+    /// @param is_first_column Whether the position is at the beginning of its line.
+    /// @return true if a document marker or a block structure indicator begins there.
+    static bool begins_non_scalar_content(str_view sv, std::size_t pos, bool is_first_column) noexcept {
+        switch (sv[pos]) {
+        case ':':
+        case '?':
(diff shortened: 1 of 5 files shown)


include/fkYAML/detail/input/lexical_analyzer.hpp
```diff
@@ -430,6 +430,46 @@ class lexical_analyzer {
...
+            break;
+        }
+
+        // Document markers are only recognized at the beginning of a line.
+        if (!is_first_column) {
+            return false;
+        }
+
+        const bool begins_marker = (sv.compare(pos, 3, "---") == 0) || (sv.compare(pos, 3, "...") == 0);
+        return begins_marker && is_followed_by_space(sv, pos + 3);
```
fktn-k: Although this should already cover ordinary use cases, I think `is_followed_by_space` should also check if `sv[pos]` is a tab according to the definition of `c-forbidden` in the specification.

fktn-k (changes requested):
I requested a minor improvement.

include/fkYAML/detail/input/lexical_analyzer.hpp
sndth: Good catch, thanks! Fixed, and it turned out to be inconsistent with the file as well as with the
spec: the lexer already accepts a tab after `:` when it decides on `KEY_SEPARATOR`, so the helper was
the odd one out.

I renamed it to `is_followed_by_white_space` since "space" was no longer accurate, matching the
wording of `skip_white_spaces()` nearby.

The tests now cover every branch of it: the document marker case runs with `---\n`, `--- ` and
`---\t`, and the value indicator case with a space, a tab, the end of the buffer and a line break.

fktn-k (approved):
The changes look good to me. Thanks!

PR #592, opened 2026-08-26
Fix anchor/alias ordering during serialization

### LINKED ISSUE #584: Serialization invalidates the anchor/alias resolving when they are specified in a mapping.
### Description

The current mapping type defaults to `std::map` which reorders its entries according to the keys. The reordering matters especially when the anchor/alias resolving is involved with mapping entries.

Imagine you deserialize the following YAML into `fkyaml::node` and serialize back to a string.
```yaml
foo: &anchor 123
bar: *anchor
```

Since the deserialized mapping reorders its entries with its keys in the alphabetic order, the serialization results in the following YAML.
```yaml
bar: *anchor
foo: &anchor 123
```

Since aliases as a `fkyaml::node` internally points to the right anchor node, the following C++ snippet works as expected.
```cpp
fkyaml::node n = fkyaml::node::deserialize("foo: &anchor\nbar: *anchor");
int i = n["bar"].get_value<int>();
assert(i == 123);
```

This happens only when `std::map` reorders its entries.  
In order to preserve the valid order, the mapping type of `fkyaml::node` should be changed to some type which preserves its insertion order, for instance, `fkyaml::ordered_map<fkyaml::node, fkyaml::node>` or `std::vector<std::pair<fkyaml::node, fkyaml::node>>`.

### Reproduction steps

Run the code shown in the minimum code example section.

Serialization output should be semantically the same as the deserialization input.  
Currently, they can be semantically different.

### Minimal code example

```Shell
#include <iostream>
#include <fkYAML/node.hpp>

int main() {
    fkyaml::node n = fkyaml::node::deserialize("foo: &anchor\nbar: *anchor");
    std::cout << n;
}
```

### Error messages

```Shell

```

GCC 13.2.0 on Ubuntu 24.04

### Library version

both v0.4.4 and the develop HEAD

### Validation

fktn-k:
This PR fixes #584, an issue where YAML documents containing anchors and aliases could not be serialized correctly after deserialization when the underlying mapping type is `std::map`, the mapping type of `fkyaml::node`.  

## Problem

`std::map` orders mapping entries by their keys, so the serialization order can differ from the original YAML document.  
This can cause a problem when an alias is serialized before the anchor it refers to. In particular, the problem an occur when:

- the same anchor name is used multiple times with different anchor occurrences;
- an anchor is redefined while traversing a sequence;
- an alias is used as a mapping key;
- the resolved value of an alias key affects the `std::map` ordering.

The deserialized node already retains the information necessary to identify which anchor occurrence an alias refers to, but the serializer did not previously that information into account when determining the mapping entry order.

## Fix

This PR adjusts the serialization order of mapping entries based on anchor/alias dependencies by stable topological sorting.  
For each mapping:

1. Anchor and alias references are collected recursively from each mapping entry.
2. Anchor occurrences are identified using the anchor name and anchor offset.
3. A mapping entry that references an anchor is postponed until the corresponding anchor has been serialized.
4. Otherwise, mapping entries retain their original `std::map` order.
5. Sequence ordering is never changed' the ordering adjustment is applied only to mapping entries.

This allows the serializer to emit anchors before the aliases that reference them while preserving the existing ordering behavior as much as possible.

I had originally considered switching to a mapping type that preserves the insertion order, but I decided against it because it would have caused significant performance degradation. Thanks to advice from @sndth in #584, I arrived at this approach, which avoids the need for a breaking change.

## Tests

Added round-trip tests to `tests/unit_test/test_serializer_class.cpp`, including cases with:

- nested mappings and sequences;
- repeated anchor names;
- anchor redefinition across entries;
- aliases used as mapping keys;
- multiple anchor/alias occurrences within the same mapping.

The tests verify that the documents can be deserialized and serialized again while preserving the correct anchor/alias relationships.

The existing serializer tests also continue to pass without any changes to their expected output.  
This confirms that the serialization output remains unchanged in cases where changing the mapping order is not necessary to preserve anchor/alias resolution.

---

## Pull Request Checklist

## Please don't

### DIFF
--- a/include/fkYAML/detail/output/serializer.hpp
+++ b/include/fkYAML/detail/output/serializer.hpp
@@ -360,7 +360,9 @@ private:
     static bool has_prior_anchor_definition(
         const std::vector<anchor_reference>& anchors, const anchor_reference& target) {
         for (const auto& anchor : anchors) {
-            if (anchor.position < target.position && is_same_anchor(anchor, target)) {
+            const bool is_anchor_defined_prior_to_target =
+                anchor.position < target.position && is_same_anchor(anchor, target);
+            if (is_anchor_defined_prior_to_target) {
                 return true;
             }
         }
@@ -430,18 +432,15 @@ private:
     /// @param items The mapping items to reorder.
     /// @param item_anchors The anchor references grouped by mapping item.
     /// @param item_aliases The alias references grouped by mapping item.
+    /// @param anchors_in_mapping The anchor references defined in the mapping.
     /// @return The mapping items in dependency-respecting serialization order.
     std::vector<map_iterator> reorder_mapping_items_by_dependencies(
         const std::vector<map_iterator>& items, const std::vector<std::vector<anchor_reference>>& item_anchors,
-        const std::vector<std::vector<anchor_reference>>& item_aliases) const {
+        const std::vector<std::vector<anchor_reference>>& item_aliases,
+        std::vector<anchor_reference> anchors_in_mapping) const {
         std::vector<bool> emitted(items.size(), false);
         std::vector<map_iterator> ordered_items;
         ordered_items.reserve(items.size());
-
-        std::vector<anchor_reference> anchors_in_mapping;
-        for (const auto& anchors : item_anchors) {
-            anchors_in_mapping.insert(anchors_in_mapping.end(), anchors.begin(), anchors.end());
-        }
 
         std::size_t emitted_count = 0;
         while (emitted_count < items.size()) {
@@ -459,21 +458,10 @@ private:
                 }
             }
 
-            if (ready_item_index == items.size()) {
-                break;
-            }
-
             emitted[ready_item_index] = true;
             ordered_items.emplace_back(items[ready_item_index]);
             remove_anchors(item_anchors[ready_item_index], anchors_in_mapping);
             ++emitted_count;
-        }
-
-        for (std::size_t i = 0; i < items.size(); ++i) {
-            const bool is_item_emitted = emitted[i];
-            if (!is_item_emitted) {
-                ordered_items.emplace_back(items[i]);
-            }
         }
 
         return ordered_items;
@@ -499,11 +487,29 @@ private:
 
         // If there are no anchors (no resolving needed) or aliases (no anchor is referenced), return the items in the
         // order `node.map_items()` returns them.
-        if (!has_any_anchor || !has_any_alias) {
+        const bool needs_reordering = has_any_anchor && has_any_alias;
+        if (!needs_reordering) {
             return items;
         }
 
-        return reorder_mapping_items_by_dependencies(items, item_anchors, item_aliases);
+        std::vector<anchor_reference> anchors_in_mapping;
+        for (const auto& anchors : item_anchors) {
+            anchors_in_mapping.insert(anchors_in_mapping.end(), anchors.begin(), anchors.end());
+        }
+
+        for (std::size_t i = 0; i < items.size(); ++i) {
+            for (const auto& alias : item_aliases[i]) {
+                const bool is_anchor_defined_in_different_item =
+                    has_anchor(anchors_in_mapping, alias) && !has_anchor(item_anchors[i], alias);
+                if (is_anchor_defined_in_different_item) {
+                    return reorder_mapping_items_by_dependencies(
+                        items, item_anchors, item_aliases, std::move(anchors_in_mapping));
+                }
+            }
+        }
+
+        // Reordering is only needed when an alias refers to an anchor in another mapping item.
+        return items;
     }
 
     void collect_anchor_alias_names(
--- a/include/fkYAML/detail/output/serializer.hpp
+++ b/include/fkYAML/detail/output/serializer.hpp
@@ -543,7 +543,8 @@ private:
     /// @return true if the mapping was serialized here, false if it needs no reordering.
     bool serialize_reordered_mapping(const BasicNodeType& node, const uint32_t cur_indent, std::string& str) {
         std::vector<map_iterator> ordered_items;
-        if (!get_mapping_items_in_serialization_order(node, ordered_items)) {
+        const bool is_reordered = get_mapping_items_in_serialization_order(node, ordered_items);
+        if (!is_reordered) {
             return false;
         }
 
--- a/include/fkYAML/detail/output/serializer.hpp
+++ b/include/fkYAML/detail/output/serializer.hpp
@@ -207,7 +207,9 @@ private:
                 return;
             }
 
-            if (m_has_anchor_table) {
+            // If there is any anchor defined for this document and the mapping has more than one entry,
+            // reorder the mapping entries so the anchor resolution order is preserved.
+            if (m_has_anchor_table && node.size() > 1) {
                 for (const auto& itr : get_mapping_items_in_serialization_order(node)) {
                     serialize_mapping_entry(itr, cur_indent, str);
                 }
--- a/include/fkYAML/detail/output/serializer.hpp
+++ b/include/fkYAML/detail/output/serializer.hpp
@@ -248,14 +248,9 @@ private:
 
             // If there is any anchor defined for this document and the mapping has more than one entry,
             // reorder the mapping entries so the anchor resolution order is preserved.
-            if (m_has_anchor_table && node.size() > 1) {
-                std::vector<map_iterator> ordered_items;
-                if (get_mapping_items_in_serialization_order(node, ordered_items)) {
-                    for (const auto& itr : ordered_items) {
-                        serialize_mapping_entry(itr, cur_indent, str);
-                    }
-                    break;
-                }
+            // Only a document which defines an anchor can ever need its mapping entries reordered.
+            if (m_has_anchor_table && node.size() > 1 && serialize_reordered_mapping(node, cur_indent, str)) {
+                break;
             }
 
             for (const auto& itr : node.map_items()) {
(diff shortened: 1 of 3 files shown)


sndth:
I kept digging after my earlier comment in #584 and ended up somewhere different from what I
suggested there. Two things: one applies to this PR as it stands, the other is about the option
you ruled out.

### 1. The early-out in `get_mapping_items_in_serialization_order()` runs after the work it avoids

`collect_anchor_alias_names()` walks the entire subtree of every mapping item, and it runs for
every mapping at every level, so a nested document is traversed once per level. `has_any_anchor` /
`has_any_alias` is only consulted once that traversal has already happened and allocated.
Documents with no anchors at all pay for it - `citm_catalog.yml` (744 KB, zero anchors),
serialization throughput:

| | serialize |
|---|---|
| `develop` | 220 MiB/s |
| this PR | 97 MiB/s |

`basic_serializer` already reaches for `node.mp_meta` when emitting directives, and
`mp_meta->anchor_table` answers "does this document have anchors" up front. Hoisting the check
there, or doing a single traversal at the root of `serialize()` instead of one per mapping, should
make this close to free for the common case. Worth doing whatever happens with the rest of this
comment.

### 2. I think the `ordered_map` measurement was hitting a bug in `ordered_map`, not a property of the approach

`ordered_map` derives from `std::vector<std::pair<const Key, Value>>`. The `const Key` member makes
the pair effectively non-movable, so every reallocation during growth copy-constructs every
entry, and for `basic_node` values that deep-copies whole subtrees, repeatedly, as the mapping
grows. That would also explain why the key-to-index map didn't help: an index removes lookups, not
copies.

Changing the base to `std::vector<std::pair<Key, Value>>` lets the entries move. Measured
on `citm_catalog.yml`:

| | parse |
|---|---|
| `develop` (`std::map`) | 51.6 MiB/s |
| default `MappingType` = `ordered_map`, as it is today | 30.3 MiB/s |
| default `MappingType` = `ordered_map` + the move fix | **55.4 MiB/s** |

The halving reproduces exactly, and it goes away - the flip ends up slightly faster than
`std::map` on that case 🫨. `ubuntu.yml` is within timer noise either way.

I've opened the `ordered_map` change on its own as a separate PR (#593), since it stands alone: anyone
already using `basic_node<std::vector, fkyaml::ordered_map>` is paying for those copies today,
regardless of what the default is.

What the default flip would buy on top of this PR is that the round trip becomes byte-identical
for documents without anchors too:

| input | `develop` | this PR | `ordered_map` default |
|---|---|---|---|
| `foo: &a 123` / `bar: *a` | does not re-parse | identical | identical |
| nested anchors | does not re-parse | identical | identical |
| repeated anchor name | does not re-parse | identical | identical |
| `zebra` / `apple` / `mango` | reordered | reordered | identical |
| `server` / `client` | reordered | reordered | identical |

and the serializer would need no reordering logic at all, since the anchor already precedes its
alias in the stored order.

Two costs I want to be upfront about, because they are real:

- **Lookup becomes O(n).** Time per `node["key"]`, `std::map` vs `ordered_map`:

  | keys | `std::map` | `ordered_map` |
  |---|---|---|
  | 10 | 98 ns | 80 ns |
  | 100 | 145 ns | 444 ns |
  | 1 000 | 214 ns | 3.9 µs |
  | 10 000 | 274 ns | 38.5 µs |

  Fine for config-sized mappings, bad for large ones. Users who need the old behaviour keep it
  with `basic_node<std::vector, std::map>` - it is one default template argument, not a removal.
- `ordered_map::value_type` changes from `std::pair<const Key, Value>` to `std::pair<Key, Value>`,
  which is a source break for code that spells the pair type out.

The default flip is entirely your call and I'm not trying to cut across this PR - if you'd rather
keep the serializer-local route, point 1 still stands on its own. I have the flip on a local
branch with tests if you want to look at the numbers yourself.

### 3. Unrelated, found while debugging this

An anchor on a block mapping doesn't reach the mapping node - it lands on the mapping's first key:

```yaml
bill-to: &id001
  given: Chris
  family: Dumars
ship-to: *id001
```

`is_anchor()` is false on the value of `bill-to`, and `*id001` resolves to the scalar `given`
instead of the mapping, so the round trip emits `&id001 given: Chris`. Block *sequences* in the
same position are handled correctly, so it looks like the "properties on a different line"
handling that exists for `SEQUENCE_BLOCK_PREFIX` has no counterpart for block mappings.
Reproduces on `develop`, and covers yaml-test-suite `26DV`, `6KGN`, `7BMT`, `U3XV`, `UGM3`.
I'll open a separate issue for it unless you'd rather have it here.

fktn-k:
@sndth  
Thanks for the detailed information!  

> **1. The early-out in get_mapping_items_in_serialization_order() runs after the work it avoids**

Oh, that should be done before merging this PR to avoid a huge serialization performance degradation. Altbough `ordered_map` can resolve the roundtrip issue, a resolution by a default mapping type doesn't cover use cases for users who want to use use a non-default type, so this PR should also be merged. Also, I've found some points which can be optimized for documents with anchors and aliases. I'll push them after some more investigation.

> **2. I think the ordered_map measurement was hitting a bug in ordered_map, not a property of the approach**

Thanks for keeping investigation on this! The performance improvement and your reasoning make sense. I just merged the PR with the breaking-change label for 0.5.0. I think the default type change should be unchanged until the O(n) lookups are optimized.  

> **3. Unrelated, found while debugging this**

Please open a separate issue since its root cause should be in the deserialization itself.

fktn-k:
@sndth  
> 1. The early-out in get_mapping_items_in_serialization_order() runs after the work it avoids

I've added three improvements regarding reordering.

1. **No reordering is done if a document has no anchors or aliases**: This implements your suggestion in a previous comment.
2. **Skip reordering for mappings with only one pair**: Obviously, there is no issue with the order of entries in such mappings.
3. **Reorder mapping entries only when an alias references an anchor in a different entry within the same mapping**: The order of mapping entries is only important when an alias refers to an anchor in a different mapping entry. If they are in the same entry, reordering is skipped, and reordering is performed in a subsequent processing if necessary.

Benchmarking serialization throughput in my local environment confirmed that the difference in performance before and after the changes introduced by this PR was negligible.

sndth:
@fktn-k 

I re-measured with the same harness as before. The early-out works: `citm_catalog.yml`, which has no anchors at all, went from 97 MiB/s to 215 MiB/s against 233 MiB/s on `develop`, so the gap is down from 56% to about 7%.

The cost did not go away though, it moved. `m_has_anchor_table` is a property of the whole document, so a single anchor anywhere re-enables the per-mapping traversal for every mapping in it:

| document | `develop` | this PR | |
|---|---|---|---|
| `citm_catalog.yml`, no anchors | 233 MiB/s | 215 MiB/s | -7% |
| 2000 small mappings, **one** anchor | 90 MiB/s | 39 MiB/s | **-57%** |
| 200 anchors + 200 aliases | 73 MiB/s | 22 MiB/s | **-70%** |

Median of 9, three alternating runs, spread under 2%.

The second row is the one I would worry about: a config with 2000 entries and one `&base` somewhere is an ordinary document, and it pays for `collect_anchor_alias_names()` walking the subtree of every mapping item, at every level.

https://github.com/fktn-k/fkYAML/blob/d1f5c6eedd240993e60697f19ec19bf67f5e2eb1/include/fkYAML/detail/output/serializer.hpp#L210-L215

Doing that collection once for the whole document, at the root of `serialize()`, would give the same information without repeating the walk per mapping: which mappings actually contain an alias that refers to an anchor in a sibling entry is a property of the tree, not of the mapping being visited.

Happy to be wrong on this if your local numbers disagree, my environment is MSVC 19.51 on Windows 11.

fktn-k:
@sndth  
Thank you for the feedback! I'm updating the implementation accordingly. I'll let you know when it's ready (hopefully soon).

fktn-k:
@sndth  
I've made modifications to minimize the cost of reordering mapping entries, even when anchors or aliases are present in a document.  

Although you suggested doing the collection once for the whole document, at the root of serialize(), since that would result in unnecessary overhead when a document consists of only sequences and scalars, I chose to cache and reuse child node information when processing each mapping.  
Furthermore, since there is no need to allocate the same anchor and alias information redundantly for each mapping, I have consolidated the actual entities into a pool and now use their indices to reference them during reordering and caching.  

With these changes, even for documents requiring anchor resolution, unnecessary recursive traversals should no longer occur, and performance degradation should be minimized.

sndth:
@fktn-k 

The early-out and the cache work: instrumenting `collect_anchor_alias_names()` on https://github.com/fktn-k/fkYAML/pull/592/changes/d191ffb954bca2b84f85b6c04b817b01460aba48 shows zero calls for `citm_catalog.yml`, and for 2000 mappings with one anchor, 16005 calls over 16005 nodes with 2000 of 2001 reorder entries hitting the cache. No repeated traversals left.

One thing remains. `reorder_mapping_items_by_dependencies()` is O(n^2) in the number of mapping entries: the inner scan restarts at index 0 after every emission, so an n-entry mapping costs about n^2/2 steps even when no entry moves.

I pushed [a branch on top of your commit](https://github.com/sndth/fkYAML/tree/speed_up_anchor_alias_reordering). Four changes:

1. Track the first unemitted index in the sort; items before it are already emitted, so skipping
   them cannot change the result.
2. Collect the events once from `serialize_document()` and cache only mappings whose subtree
   produced an event, so a mapping missing from the cache returns immediately with no event scan
   and no item vector. `cache_mapping` and `p_mapping_reference` become unnecessary.
3. Keep the reordering in its own function. Declaring the item vector directly in `serialize_node()`
   costs about 1% on documents without anchors even though the branch never runs there, since it is
   called for every node.
4. `is_same_anchor()` compares the offset before the name.

Release (MSVC 19.51), fastest of 11 alternating runs:

| document | `develop` | `d191ffb` | branch |
|---|---|---|---|
| `citm_catalog.yml`, no anchors | 234.6 MiB/s | 225.0 | 227.7 |
| `ubuntu.yml`, no anchors | 223.3 MiB/s | 216.3 | 216.9 |
| 200 anchors + 200 aliases | 31.9 MiB/s | 12.3 | 12.5 |
| 2000 mappings, one anchor | 90.2 MiB/s | 47.7 | 58.6 |
| 2000 mappings, anchor + alias | 90.1 MiB/s | 24.6 | 57.0 |
| 2000 mappings, reordering needed | 89.9 MiB/s | 46.9 | 57.1 |

Scaling an N-entry root mapping holding `base: &base` and `uses: *base`:

| entries | `develop` | `d191ffb` | branch |
|---|---|---|---|
| 250 | 0.34 us/entry | 0.72 | 0.52 |
| 1000 | 0.34 us/entry | 0.97 | 0.55 |
| 4000 | 0.34 us/entry | 1.89 | 0.55 |

Output is byte-identical to your branch, with 390/390 unit tests and the yaml-test-suite result unchanged. Take it or leave it, whichever is easier 😸

fktn-k:
@sndth  
Thanks for the feedback! I `cherry-pick`ed your commits with an additional commit to improve coverage.  
I will merge this PR since the initial issue and major performance concerns are resolved.  
Please open a PR if you find anything which can improve this feature. I'd be happy to accept it.

PR #593, opened 2026-08-26
Let ordered_map entries be moved when the storage grows

sndth:
`ordered_map` derives from `std::vector<std::pair<const Key, Value>>`. The `const Key` member makes the pair effectively non-movable, so every reallocation during growth copy-constructs every entry instead of moving it. For `basic_node` values that means deep-copying whole subtrees, repeatedly, as the mapping grows.

This changes the base container and the default `Allocator` to `std::pair<Key, Value>` so the entries move.

### Effect

`ordered_map` is not the default `MappingType`, so this does not change anything for `fkyaml::node`. It does change things for anyone using `basic_node<std::vector, fkyaml::ordered_map>` today, and it is measurable: flipping the default locally (not part of this PR) and running the repo benchmarker on `citm_catalog.yml` in Release:

| | parse |
|---|---|
| `develop`, `std::map` | 51.6 MiB/s |
| `ordered_map` before this change | 30.3 MiB/s |
| `ordered_map` after this change | 55.4 MiB/s |

This is relevant to #584 / #592: the ~2x slowdown reported there when trying `ordered_map` as the default reproduces exactly, and it comes from these copies rather than from the linear `find()`. That is also why adding a key-to-index map did not help - an index removes lookups, not copies. Whether the default `MappingType` should change is a separate question and is not proposed here.

### Breaking change

`ordered_map::value_type` changes from `std::pair<const Key, Value>` to `std::pair<Key, Value>`. Code that spells the pair type out explicitly, e.g.

```cpp
for (std::pair<const std::string, int>& entry : map) { ... }
```

needs the `const` dropped. Iterating with `auto&` is unaffected. Keys become assignable in place, which callers should not do - the class has no way to detect the duplicates that would create.

### Tests

- `OrderedMap_InsertionOrderSurvivesGrowth`: insertion order across the reallocations that growth
  causes.
- `OrderedMap_AsMappingTypeKeepsKeyOrder`: round trip of the #584 document through
  `basic_node<std::vector, fkyaml::ordered_map>`.

Both the unit tests and the yaml-test-suite runner pass unchanged on this branch.

---

## Pull Request Checklist

## Please don't

### DIFF
--- a/docs/docs/api/ordered_map/index.md
+++ b/docs/docs/api/ordered_map/index.md
@@ -5,12 +5,12 @@
 ```cpp
 template<
     typename Key, typename Value, typename IgnoredCompare = std::less<Key>,
-    typename Allocator = std::allocator<std::pair<const Key, Value>>>
-class ordered_map : public std::vector<std::pair<const Key, Value>>;
+    typename Allocator = std::allocator<std::pair<Key, Value>>>
+class ordered_map : public std::vector<std::pair<Key, Value>>;
 ```
 
 A minimal map-like container which preserves insertion order.  
-This documentation only describes APIs which are not of the parent class [`std::vector<std::pair<const Key, Value>>`](https://en.cppreference.com/w/cpp/container/vector).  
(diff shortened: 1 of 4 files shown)


fktn-k (approved):
LGTM. Thanks!

PR #595, opened 2026-08-29
Count a document only if it has contents or a directives end marker

sndth:
`deserialize_docs()` emits a node for every stream section, including one which holds only comments, white spaces or a bare `...`. Such a section is not a document, so an empty node is added for it.

This counts a document only if it has contents or begins with an explicit `---`. A `---` which ends a document still begins the next one, so a trailing marker keeps producing an empty document.

Refs #591.

### Results

yaml-test-suite: 95 -> 84 failures with no regressions. `8G76`, `98YD`, `AVM7`, `HWV9`, `M7A3` and `QT73` now pass.

### Not covered

Consecutive directives end markers are still counted as a single document, so `6XDY` keeps failing:

```yaml
---
---
```

`deserialize_docs()` returns 1 document instead of 2, because `deserialize_directives()` consumes every `---` it sees within one call. Changing that looked separable from this PR, so I left it out. Happy to fold it in here if you would rather have both.

---

## Pull Request Checklist

## Please don't

### DIFF
--- a/include/fkYAML/detail/input/deserializer.hpp
+++ b/include/fkYAML/detail/input/deserializer.hpp
@@ -158,9 +158,15 @@ public:
 
         std::vector<basic_node_type> nodes {};
         lexical_token_t type {lexical_token_t::END_OF_BUFFER};
+        bool begun_by_marker = false;
 
         do {
-            nodes.emplace_back(deserialize_document(lexer, type));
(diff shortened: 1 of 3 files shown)


fktn-k (approved):
LGTM. Thanks a lot!

PR #597, opened 2026-08-29
Support a flow mapping entry which omits the value indicator

sndth:
A flow mapping entry may consist of a key alone, with both the `:` indicator and the value omitted, but such an entry is rejected with `The ":" mapping value indicator must be followed after a mapping key.`

```yaml
{foo, bar: baz}
```

`ns-flow-map-yaml-key-entry` allows an `e-node` in place of the separate value, so the entry ends at the `,` separator or at the `}` suffix which follows the key.

A key followed by anything else is still rejected, and the relaxation does not reach the block context, where `,` and `}` remain errors.

### Results

yaml-test-suite: 95 -> 92 failures with no regressions. `4ABK`, `8KB6` and `9BXH` now pass.

---

## Pull Request Checklist

## Please don't


fktn-k:
@sndth Thanks for the contribution! I'll take a look later.

fktn-k (approved):
LGTM. Thanks a lot!

PR #598, opened 2026-08-30
Fix tab used as indentation

sndth:
Indentation must consist of spaces, while a tab which follows it is valid separation white space. Tabs used where indentation is required are currently accepted:

```yaml
-	-              # a nested sequence indented with a tab
?	key:           # a mapping used as an explicit key
foo: "bar
	baz"           # a continuation line of a quoted scalar
- [
	foo, bar ]     # a continuation line of a flow collection
```

Tabs which only separate tokens keep working:

```yaml
-	-1             # separates the entry from its scalar
foo:
 	bar            # the space indents, the tab follows
"1st
	2nd"           # the scalar begins the line, so it owns that indentation
```

Three places consult the required indentation: the deserializer asks the lexer wherever a block indicator begins a collection, the quoted scalar scanners check their own continuation lines, and the lexer checks each new line within a flow collection. Nothing is tracked while lexing documents which contain no tabs.

`?` also accepts a tab after it now, the same way `-` already did. Without it `?\t` was not an explicit key indicator at all.

Refs https://github.com/fktn-k/fkYAML/discussions/591 (first group).

### Results

yaml-test-suite: 78 -> 70 failures with no regressions. `4EJS`, `DK95-01`, `Y79Y-003`, `Y79Y-004`, `Y79Y-005`, `Y79Y-006`, `Y79Y-007` and `Y79Y-008` now pass, which covers every case the suite marks invalid because of a tab. The 45 valid cases which contain tabs are unaffected.

---

## Pull Request Checklist

## Please don't

### DIFF
--- a/include/fkYAML/detail/input/deserializer.hpp
+++ b/include/fkYAML/detail/input/deserializer.hpp
@@ -218,6 +218,7 @@ private:
 
         switch (token.type) {
         case lexical_token_t::SEQUENCE_BLOCK_PREFIX: {
+            check_tab_in_indentation(lexer, lexer.get_lines_processed(), lexer.get_last_token_begin_pos());
             root = basic_node_type::sequence({basic_node_type()});
             apply_directive_set(root);
             if (found_props) {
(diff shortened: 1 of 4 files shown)


fktn-k (approved):
LGTM. `lexer::check_continuation_indent(str_view content)` may affect parsing performance, but this only applies when the target scalar spans multiple lines and the content of each line is long. So, I'm fine to leave it as-is for now. In the future, we may be able to optimize the scanning logic.

Issue #583, opened 2026-08-23
Libraries' versions used for the benchmark are outdated


fktn-k:
### Description

(This is not actually an issue in fkYAML but rather a note for myself)

The benchmark uses libfyaml, rapidyaml and yaml-cpp.  
However, they have released newer versions and thus the result is not up-to-date.  
The benchmark should be taken with their latest versions before the next release (v0.5.0).

These are the latest versions available as of 2026/8/23.
* libfyaml: 1.0.0-beta1
* rapidyaml: 0.16.0
* yaml-cpp: 0.9.0

### Reproduction steps

N/A

### Expected vs. actual results

Expected: The latest versions of the libraries should be used.
Actual: Some outdated versions are used.

### Minimal code example

```Shell

```

### Error messages

```Shell

```

### Compiler and operating system

N/A

### Library version

v0.4.4

### Validation

- [ ] The bug also occurs if the latest version from the [`develop`](https://github.com/fktn-k/fkYAML/tree/develop) branch is used.
- [ ] I can successfully [compile and run the unit tests](https://github.com/fktn-k/fkYAML#how-to-execute-the-unit-tests).

fktn-k:
The libraries have been updated by the PR #603.

PR #603, opened 2026-08-31
Update the YAML libraries used for the benchmark

sndth:
This PR updates libraries used for the benchmark to the latest available:

| library | before | after |
|---|---|---|
| libfyaml | `v0.9` | `v1.0.0-beta1` |
| rapidyaml | `v0.7.2` | `v0.16.0` |
| yaml-cpp | `0.8.0` | `yaml-cpp-0.9.0` |

Two things the bump needed beyond the tags:

- rapidyaml no longer uses submodules as of 0.16.0, so `GIT_SUBMODULES "ext/c4core"` is removed.
  Keeping it fails the fetch.
- yaml-cpp changed its tag naming. `0.8.0` is a bare tag, but 0.9.0 exists only as `yaml-cpp-0.9.0`.

libfyaml supports Windows natively since 0.9.4, so the `if(NOT WIN32)` guard is gone and it is now
benchmarked there too. That needed `set(BUILD_TESTING OFF)`, since its tests download the YAML and
JSON test suites at configure time and cloning the latter fails on the Windows path length limit,
and `set(BUILD_SHARED_LIBS OFF)`, since libfyaml defaults that option to `ON` for every dependency
and the resulting DLLs are not found next to `benchmarker.exe`.

I have not regenerated `tools/benchmark/results/`, since those were measured on your machine and a
rerun here would just swap the hardware. They are worth regenerating on the reference environment
before the release.

Refs #583.

---

## Pull Request Checklist

## Please don't

### DIFF
--- a/tools/benchmark/CMakeLists.txt
+++ b/tools/benchmark/CMakeLists.txt
@@ -11,7 +11,7 @@ set(BENCHMARK_ENABLE_TESTING OFF CACHE BOOL "" FORCE)
 set(BENCHMARK_ENABLE_GTEST_TESTS OFF CACHE BOOL "" FORCE)
 FetchContent_Declare(
   gbench
-  URL https://github.com/google/benchmark/archive/refs/tags/v1.8.4.zip
+  URL https://github.com/google/benchmark/archive/refs/tags/v1.9.5.zip
 )
 FetchContent_MakeAvailable(gbench)
 
@@ -24,7 +24,7 @@ if(NOT WIN32)
   FetchContent_Declare(
     libfyaml
     GIT_REPOSITORY https://github.com/pantoniou/libfyaml
-    GIT_TAG v0.9
+    GIT_TAG v1.0.0-beta1
   )
   FetchContent_MakeAvailable(libfyaml)
 endif()
@@ -33,8 +33,7 @@ endif()
 FetchContent_Declare(
   rapidyaml
   GIT_REPOSITORY https://github.com/biojppm/rapidyaml
-  GIT_TAG v0.7.2
-  GIT_SUBMODULES "ext/c4core"
+  GIT_TAG v0.16.0
 )
 FetchContent_MakeAvailable(rapidyaml)
 
@@ -44,7 +43,7 @@ set(YAML_CPP_BUILD_TOOLS OFF CACHE BOOL "" FORCE)
 FetchContent_Declare(
   yaml_cpp
   GIT_REPOSITORY https://github.com/jbeder/yaml-cpp
-  GIT_TAG 0.8.0
+  GIT_TAG yaml-cpp-0.9.0
 )
 FetchContent_MakeAvailable(yaml_cpp)
 
--- a/tools/benchmark/CMakeLists.txt
+++ b/tools/benchmark/CMakeLists.txt
@@ -19,15 +19,17 @@ FetchContent_MakeAvailable(gbench)
 #   Set up YAML parsers   #
 ###########################
 
-# libfyaml (not implemented for Windows)
-if(NOT WIN32)
-  FetchContent_Declare(
-    libfyaml
-    GIT_REPOSITORY https://github.com/pantoniou/libfyaml
-    GIT_TAG v1.0.0-beta1
-  )
-  FetchContent_MakeAvailable(libfyaml)
-endif()
+# libfyaml
+# build statically so that no shared library has to be found at run time, and disable building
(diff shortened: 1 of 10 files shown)


tools/benchmark/CMakeLists.txt
```diff
@@ -24,7 +24,7 @@ if(NOT WIN32)
   FetchContent_Declare(
     libfyaml
     GIT_REPOSITORY https://github.com/pantoniou/libfyaml
-    GIT_TAG v0.9
+    GIT_TAG v1.0.0-beta1
```
fktn-k: It seems that libfyaml has added full native Windows support since 0.9.4. So it would be nice if you remove the `if(NOT WIN32)` guard.  
https://github.com/pantoniou/libfyaml/releases/tag/v0.9.4

fktn-k (changes requested):

tools/benchmark/CMakeLists.txt
sndth: I must have missed the update where libfyaml introduced full Windows support. Fixed, thanks!

Two settings were needed to make that work:

- `set(BUILD_TESTING OFF)`: libfyaml downloads the YAML and JSON test suites at configure time to run its own tests. Cloning the latter fails on Windows on the path length limit, so the configure step never finishes.
- `set(BUILD_SHARED_LIBS OFF)`: libfyaml defaults that option to `ON`, which then applies to rapidyaml and yaml-cpp as well. The resulting DLLs do not end up next to `benchmarker.exe`, so it exits with `STATUS_DLL_NOT_FOUND`.

`FK_YAML_BM_HAS_LIBFYAML` is removed too, since it is now always defined.

I have also reverted the Google Benchmark bump. You pinned v1.8.4 as the latest version supporting C++11, and this PR is about the YAML libraries, so changing it does not belong here. For what it is worth, the constraint turned out to be narrower than the note suggests: v1.9.5 sets `BENCHMARK_CXX_STANDARD 17` only for its own sources, `benchmark::benchmark` exports no `cxx_std_*` requirement, and `benchmark.h` still keeps its `__cplusplus < 201402L` fallback. So only the toolchain building Google Benchmark itself would need C++17, not the benchmarking app. Happy to open a separate issue if you want it updated.

Measured on Windows (MSVC 19.51, Release, `citm_catalog.yml`):

| benchmark | throughput |
|---|---|
| `bm_fkyaml_parse` | 53.5 MiB/s |
| `bm_yamlcpp_parse` | 8.0 MiB/s |
| `bm_libfyaml_parse` | 35.0 MiB/s |
| `bm_rapidyaml_parse_inplace` | 143.5 MiB/s |
| `bm_rapidyaml_parse_arena` | 136.3 MiB/s |

`tools/benchmark/README.md` is updated to match the new pins as well.

fktn-k: The changes look good to me. Thanks!

fktn-k:
@sndth  
Thank you for investigating the release notes of Google Benchmark.  
https://github.com/fktn-k/fkYAML/pull/603#discussion_r3906759032  

It would be nice to bump its version to 1.9.5 if it's still compatible with C++11! I would appreciate it if you open a PR for it.  

I will regenerate the benchmark results this weekend since the PC I use for benchmarking isn't available right now.   
As before, it would be helpful if you could include the regenerated results in this PR. If that's difficult, I'm happy to merge this and open a separate PR.

fktn-k (approved):

sndth:
> @sndth Thank you for investigating the release notes of Google Benchmark. [#603 (comment)](https://github.com/fktn-k/fkYAML/pull/603#discussion_r3906759032)
> 
> It would be nice to bump its version to 1.9.5 if it's still compatible with C++11! I would appreciate it if you open a PR for it.
> 
> I will regenerate the benchmark results this weekend since the PC I use for benchmarking isn't available right now. As before, it would be helpful if you could include the regenerated results in this PR. If that's difficult, I'm happy to merge this and open a separate PR.

Thank you for reply! It's done.

fktn-k:
@sndth  
I regenerated benchmark results and pushed a commit to the [update_benchmark_result](https://github.com/fktn-k/fkYAML/tree/update_benchmark_result) branch. I would appreciate it if you could incorporate it to this PR.

sndth:
@fktn-k 

Sorry for late response and thanks! Your commit is now in this PR, so it's ready to merge from my side 😺

fktn-k:
@sndth  
No worries! Thanks a lot for your help!

PR #604, opened 2026-08-31
Bind node properties to the node they precede

### LINKED ISSUE #594: An anchor for a block mapping is applied to its first key
### Description

An anchor which precedes a block mapping is applied to the first key of that mapping instead of
to the mapping itself. The mapping is then not an anchor node, and an alias which refers to that
anchor resolves to the key scalar. A block sequence in the same position is handled correctly.
This also breaks the round trip, since serializing the result emits the anchor on the key.

The yaml-test-suite cases `26DV`, `6KGN`, `7BMT`, `U3XV` and `UGM3` fail for this reason.

### Reproduction steps

Parse the following YAML:

```yaml
foo: &anchor
  bar: baz
qux: *anchor
```

`&anchor` is expected to be applied to the mapping `{bar: baz}`, so that `root["foo"].is_anchor()`
returns true and `*anchor` resolves to that mapping. Instead the anchor is applied to the key
`bar`: `root["foo"].is_anchor()` returns false, `*anchor` resolves to the string `bar`, and the
serialized output is a different document from the input.

### Minimal code example

```cpp
#include <iostream>
#include <fkYAML/node.hpp>

int main() {
    auto root = fkyaml::node::deserialize("foo: &anchor\n  bar: baz\nqux: *anchor\n");

    std::cout << root["foo"].is_anchor() << std::endl;
    std::cout << fkyaml::to_string(root["qux"].get_type()) << std::endl;
    std::cout << fkyaml::node::serialize(root);
}

// output:
// 0
// STRING
// foo:
//   &anchor bar: baz
// qux: *anchor
```

### Error messages

```Shell

```

MSVC, Windows 11

### Library version

0.4.4

### Validation

sndth:
This PR fixes #594, node properties which precede their node by a line are bound to the mapping key which follows them instead of to the node they belong to.

The properties are applied before it is known whether the scalar which follows is a value or a mapping key, so they end up on the key:

```yaml
foo: &anchor
  bar: baz     # the anchor lands on "bar" instead of on the mapping
- !circle
  center: 1    # parse_error: Detected invalid indentation.
```

## What's changed

* The binding is deferred until the token after the scalar decides what it is, which is the point suggested in #594. If a key separator follows, the properties belong to the mapping which the key begins, otherwise the node is a value and keeps them.

* Properties which precede their node are held apart from those found on the node's own line, so both reach the right node. Sharing one slot meant the later ones overwrote the earlier:

```yaml
top: &map
  &key bar: baz   # &map is for the mapping, &key is for the "bar" key
```

* This makes the special case for a mapping tag unnecessary, so it is removed. It existed to apply the tag before the first key could overwrite it, and both now survive instead:

```yaml
foo: !!map
  !!str 123: true   # !!map stays on the mapping, !!str goes to the key
```

* A deferred tag no longer reaches `resolve_scalar_tag()`, so a collection tag is not rejected as a scalar one.

* An alias node cannot carry deferred properties, which is only known once it turns out not to be a key of the mapping they belong to.

* If nothing follows them but the next entry, the properties belong to the empty value they preceded:

```yaml
foo: &anchor
bar: 1        # the anchor is for the empty value of "foo"
```

Properties on the same line as their node keep binding to it, and a block sequence on the next line is unaffected.

All unit tests pass.
This also fixes `26DV`, `7BMT`, `C4HZ`, `U3XV` and `UGM3` in the YAML test suite, as well as `6JWB`, `735Y` and `BU8L`, which failed on `A sequence or mapping tag cannot be specified to a scalar node`.
As a result, 68 failing cases has decreased to 60 with no newly failing case.

`6KGN` and `9KAX` still fail, for reasons which are not about which node the properties are bound to:

* `6KGN` needs an anchor which is still deferred to be visible to alias resolution. The anchor does reach the empty value it belongs to, but an alias on the next line is looked up before it is registered.
* `9KAX` is about properties which precede the document root node, which has a path of its own rather than the one this PR changes.

---

## Pull Request Checklist

## Please don't

### DIFF
--- a/include/fkYAML/detail/input/deserializer.hpp
+++ b/include/fkYAML/detail/input/deserializer.hpp
@@ -661,27 +661,6 @@ private:
                         add_explicit_key_with_empty_value(old_line, old_indent);
                     }
 
-                    if (m_needs_tag_impl) {
-                        const tag_t tag_type = tag_resolver_type::resolve_tag(m_tag_name, mp_meta);
-                        if (tag_type == tag_t::MAPPING || tag_type == tag_t::CUSTOM_TAG) {
-                            // set YAML node properties here to distinguish them from those for the first key node
-                            // as shown in the following snippet:
-                            //
-                            // ```yaml
-                            // foo: !!map
-                            //   !!str 123: true
-                            //   ^
-                            //   this !!str tag overwrites the preceding !!map tag.
-                            // ```
-                            *mp_current_node = basic_node_type::mapping();
-                            apply_directive_set(*mp_current_node);
-                            apply_deferred_properties(*mp_current_node);
-                            apply_node_properties(*mp_current_node);
-                            m_context_stack.emplace_back(line, indent, context_state_t::BLOCK_MAPPING, mp_current_node);
-                            continue;
-                        }
-                    }
-
                     if (token.type == lexical_token_t::SEQUENCE_BLOCK_PREFIX) {
                         // a key separator preceding block sequence entries
                         *mp_current_node = basic_node_type::sequence({basic_node_type()});
--- a/include/fkYAML/detail/input/deserializer.hpp
+++ b/include/fkYAML/detail/input/deserializer.hpp
@@ -224,6 +224,7 @@ private:
             if (found_props) {
                 // If node properties are found before the block sequence entry prefix, the properties belong to the
                 // root sequence node.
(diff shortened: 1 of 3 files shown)


fktn-k (approved):
@sndth  
LGTM. Thanks a lot for the great contribution!

PR #605, opened 2026-08-31
Let any number of white spaces separate plain scalar characters

sndth:
This PR fixes a plain scalar being cut short at the first pair of consecutive white spaces, so ordinary documents fail to parse:

```cpp
auto root = fkyaml::node::deserialize("foo: bar  baz\n");
// parse_error: The ":" mapping value indicator must be followed after a mapping key.
```

The clearest sign that this is not intended is that one space and two spaces disagree. On `develop`:

| input | result |
|---|---|
| `test: bar ]` | `bar ]` |
| `test: bar  ]` | `parse_error: Flow sequence ending is found outside the flow context.` |
| `test: a ? b` | `a ? b` |
| `test: a  ? b` | parses, but the value is silently truncated to `a` |

The last row loses data without an error.

The specification allows any amount of white space between the characters of a plain scalar, and in a block context none of the indicators are special once the scalar has begun:

```
[134] nb-ns-plain-in-line(c) ::= ( s-white* ns-plain-char(c) )*
[132] ns-plain-safe(FLOW-OUT) ::= ns-char
```

https://yaml.org/spec/1.2.2/#733-plain-style

## What's changed

* When white space is found, the whole run of it is skipped and the character which follows the run decides whether the scalar ends, instead of only the character right after the first space.
* What ended a plain scalar before still ends it: the comment prefix ` #`, a line break, ` : `, and a flow indicator in a flow context. Trailing white space is still trimmed.

## Tests which described the old behaviour

Seven inputs asserted results which only occurred because of this, so they are updated:

* `test: bar  ]` and `test: foo  }` expected an error. Both are plain scalars in a block context.
* `a  ? a` in the fuzz regressions expected an error. It is the plain scalar `a  ? a`.
* A lexer case expected `|-` after contents to begin a block scalar header. A header only begins a block scalar where a node does, so it is an ordinary character there.
* Three "missing value separators" inputs used plain scalars, where white space cannot separate entries at all. They are replaced with inputs where the next entry cannot continue the current one, such as `[[1] [2]]` and `["a" "b"]`.

The two inputs which were commented out in that same case, `[123  [true, 3.14]]` and `[123  {foo: true, bar: 3.14}]`, work as error cases now and are enabled again.

New tests cover two or more spaces, tabs, indicators which follow them, the flow context, and each way a plain scalar still ends.

The YAML test suite is unchanged at 68 failing cases, since it contains no case with a plain scalar which has consecutive white spaces inside it. That is presumably why this survived: I ran into it through [throneproj/Throne#1338](https://github.com/throneproj/Throne/issues/1338), where Clash subscription names such as `Cool  Name` fail to parse while other clients accept them.

---

## Pull Request Checklist

## Please don't

### DIFF
--- a/include/fkYAML/detail/input/lexical_analyzer.hpp
+++ b/include/fkYAML/detail/input/lexical_analyzer.hpp
@@ -1162,39 +1162,40 @@ private:
                 break;
             }
             case ' ':
-            case '\t':
-                if FK_YAML_UNLIKELY (pos == sv.size() - 1) {
-                    // trim trailing space.
+            case '\t': {
+                // Any number of white spaces may separate the characters of a plain scalar, so the whole
+                // run of them belongs to it and what follows the run decides whether it ends.
+                // See https://yaml.org/spec/1.2.2/#733-plain-style for more details.
+                const std::size_t next_pos = sv.find_first_not_of(" \t", pos + 1);
+                if FK_YAML_UNLIKELY (next_pos == str_view::npos) {
+                    // trim trailing white space.
                     ends_loop = true;
                     break;
                 }
 
-                // Allow a space in a plain scalar only if the space is surrounded by non-space characters, but not
-                // followed by the comment prefix " #".
-                // Also, flow indicators are not allowed to be followed after a space in a flow context.
-                // See https://yaml.org/spec/1.2.2/#733-plain-style for more details.
-                switch (sv[pos + 1]) {
-                case ' ':
-                case '\t':
+                // White space is not allowed to be followed by the comment prefix " #", and flow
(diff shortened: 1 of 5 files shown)


fktn-k (approved):
LGTM. I actually didn't know such a syntax is also valid. Since I couldn't find any example in the specification and the YAML test suite, I confirmed the output with [the YAML Playground](https://play.yaml.com/?show=JlpUgfxrsSdLnPhH#), which internally uses an official reference parser.

fktn-k:
@sndth  
Please resolve the conflict. Thank you.

sndth:
> @sndth Please resolve the conflict. Thank you.

It's done!

PR #606, opened 2026-09-03
Update Google Benchmark to v1.9.5

sndth:
This PR updates the Google Benchmark library used for the benchmark from `v1.8.4` to `v1.9.5` (requested in #603).

The note calling `v1.8.4` the latest version supporting C++11 turned out to be narrower than it reads. Google Benchmark's own README for v1.9.5 states that the library can be used with C++11 but requires C++17 to build, including compiler and standard library support. That holds up in the code:

- `benchmark::benchmark` exports no `cxx_std_*` compile feature, and `BENCHMARK_CXX_STANDARD 17` is
  a plain variable set in Google Benchmark's own directory scope, so nothing is imposed on the
  `benchmarker` target.
- `benchmark.h` uses nothing newer than C++11 and still keeps its `__cplusplus < 201402L` fallback
  for `make_unique`.

So fkYAML and `tools/benchmark/main.cpp` keep compiling as before. What changes is that the toolchain building the benchmarking tool now needs C++17 support. That applies to this tool only, not to the library or its supported compilers, and the tool is not built in CI.

Verified on Windows (MSVC 19.51).

---

## Pull Request Checklist

## Please don't

### DIFF
--- a/tools/benchmark/CMakeLists.txt
+++ b/tools/benchmark/CMakeLists.txt
@@ -11,7 +11,7 @@ set(BENCHMARK_ENABLE_TESTING OFF CACHE BOOL "" FORCE)
 set(BENCHMARK_ENABLE_GTEST_TESTS OFF CACHE BOOL "" FORCE)
 FetchContent_Declare(
   gbench
-  URL https://github.com/google/benchmark/archive/refs/tags/v1.8.4.zip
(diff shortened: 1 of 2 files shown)


fktn-k (approved):
I will update the benchmark results in the PR #603.

PR #610, opened 2026-09-07
Fix a block scalar whose first content line holds a single character

sndth:
This PR fixes a block scalar whose first content line consists of a single character. Such a scalar absorbs the line which follows it, so the contents of that line are lost and the scalar itself gets a wrong value:

```yaml
foo: |
  a
bar: 1
```

The above input is deserialized into `{foo: "a\nr: 1\n"}` and the `bar` entry disappears entirely. The same happens in a block sequence, where:

```yaml
- |
  x
- 2
```

yields `["x\n2\n"]` instead of two entries. A block scalar with two or more content lines is not affected, which is why this has gone unnoticed so far.

## Changes

* Fixed the search for the newline which ends the first content line in `determine_block_scalar_content_range()`. The search started one character after the cursor, which already points past the first character of that line, so the newline was skipped whenever the line holds just a single character. The scan then continued into the following line and the block scalar absorbed it.

## Testing

* Added a unit test to `test_deserializer_class.cpp` which verifies that a block scalar with a single character on its first content line ends at that line and that the following entry is preserved.
* `KK5P` and `M6YH` in the YAML test suite are now deserialized correctly. As a result, 53 failing cases has decreased to 51 with no newly failing case.

---

## Pull Request Checklist

## Please don't

### DIFF
--- a/include/fkYAML/detail/input/lexical_analyzer.hpp
+++ b/include/fkYAML/detail/input/lexical_analyzer.hpp
@@ -1340,7 +1340,9 @@ private:
             emit_error("The first non-empty line in the block scalar is less indented.");
         }
 
-        std::size_t last_newline_pos = sv.find('\n', cur_itr - m_token_begin_itr + 1);
+        // cur_itr already points past the first character of the first non-empty line, so the newline
+        // which ends that line is at that very position when the line holds a single character.
+        std::size_t last_newline_pos = sv.find('\n', cur_itr - m_token_begin_itr);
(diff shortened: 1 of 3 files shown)


fktn-k (approved):
The changes look good to me.  
Although some CI jobs are failing at installing git and unzip for older gcc/clang compilers, I'll merge this since those for modern gcc/clang compilers pass and the error logs seem to be caused by cache in the runner images.

PR #611, opened 2026-09-07
End a plain scalar at flow separator

sndth:
This PR is a follow up to #610 and fixes another case where a scalar is not terminated where it should be. A plain scalar in a flow context is not terminated by a value separator (`,`) beginning the line that follows it, so it absorbs the separator and the entries after it and the flow collection loses its structure:

```yaml
[ a
, b ]
```

The above input is deserialized into `["a , b"]`, a sequence of one entry, instead of `["a", "b"]`. The same shape in a flow mapping:

```yaml
{ a: 1
, b: 2 }
```

is rejected with `Detected invalid indentation.`, since the plain scalar `1 , b` leaves the following `:` without a key. Beginning a continuation line with the separator is a common way to format multi-line flow collections and is valid YAML in both cases.

The cause is in `determine_plain_scalar_range()`. When the scan reaches a line break it moves to the first non-space character of the continuation line, but the loop then advances past that character with `find_first_of(filter, pos + 1)` before the switch inspects it. The only place that character is examined is the `begins_non_scalar_content()` call in the line break handler, which recognized `:`, `?` and the flow collection delimiters but not the value separator.

## Changes

* Added `,` to `begins_non_scalar_content()` so that a value separator which begins a line ends the preceding plain scalar in a flow context, as required by [the YAML specification](https://yaml.org/spec/1.2.2/#733-plain-style). In a block context a comma is an ordinary plain scalar character and is left as it is.
* `begins_non_scalar_content()` is no longer static, as the new check depends on whether the lexer is currently in a flow context.

## Testing

* Added a unit test to `test_deserializer_class.cpp` which verifies that a value separator beginning a line separates the entries of a flow mapping.
* `LP6E` in the YAML test suite is now deserialized correctly. As a result, 53 failing cases has decreased to 52 with no newly failing case.

---

## Pull Request Checklist

## Please don't

### DIFF
--- a/include/fkYAML/detail/input/lexical_analyzer.hpp
+++ b/include/fkYAML/detail/input/lexical_analyzer.hpp
@@ -459,11 +459,12 @@ private:
     }
 
     /// @brief Checks if the given position begins something a plain scalar cannot continue into.
+    /// @note Must not be static, as flow indicators only end a plain scalar within a flow context.
     /// @param sv The buffer contents being scanned.
     /// @param pos The position of the first non-space character in a line.
     /// @param is_first_column Whether the position is at the beginning of its line.
     /// @return true if a document marker or a block structure indicator begins there.
-    static bool begins_non_scalar_content(str_view sv, std::size_t pos, bool is_first_column) noexcept {
+    bool begins_non_scalar_content(str_view sv, std::size_t pos, bool is_first_column) const noexcept {
         switch (sv[pos]) {
         case ':':
         case '?':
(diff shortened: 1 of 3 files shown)


fktn-k (approved):
The changes loik good to me. Thanks a lot!  
Regarding the workflow failures, there seems to be some external cause for which wcan do nothing...  
https://discuss.streamlit.io/t/release-file-for-http-deb-debian-org-debian-security-dists-bullseye-security-inrelease-is-expired-invalid-since-8h-51min-41s/122366

PR #617, opened 2026-09-09
Fix trailing white spaces not being trimmed when folding flow scalars

sndth:
This PR fixes trailing white spaces before a line break not being discarded in multiline flow scalars:

```yaml
"2 trailing\t  
    tab"
```

is parsed as `2 trailing\t   tab` (three spaces) instead of `2 trailing\t tab`. Plain and single quoted scalars are affected the same way.

### Cause

The trimming in `process_line_folding()` relies on `str_view::find_last_not_of()`, which never matched. That overload and `find_last_of()` both reuse `rfind()`'s logic, where `n` is the length of the sequence being searched for, but here `n` is the size of the character set, so `std::min(m_len - n, pos)` started the search `n - 1` characters before the end and skipped the last character. `find_last_of()` also read out of bounds when the set was as long as the string (it has no caller in the library today).

Both now start at the last character, and the trimming range stops in front of the line break instead of covering it, otherwise a working search would have kept the break itself in the scalar.

### Tests

`StrView_FindLastOf` no longer passes the string size as the character set size, which read past the literal it searched, and gains an empty string and an empty set. Three more YAML test suite cases pass: `DE56-01`, `DE56-03`, `NAT4`.

---

## Pull Request Checklist

## Please don't

### DIFF
--- a/include/fkYAML/detail/input/scalar_parser.hpp
+++ b/include/fkYAML/detail/input/scalar_parser.hpp
@@ -433,11 +433,8 @@ private:
     /// @param newline_pos Position of the target newline code.
     void process_line_folding(str_view& token, std::size_t newline_pos) noexcept {
         // discard trailing white spaces which precedes the line break in the current line.
-        const std::size_t last_non_space_pos = token.substr(0, newline_pos + 1).find_last_not_of(" \t");
-        if (last_non_space_pos == str_view::npos) {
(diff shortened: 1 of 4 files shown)


fktn-k (approved):
Good catch! Thank you for the fix.  
I'll merge tbis PR since the changes look good to me😊

PR #619, opened 2026-09-10
Fix empty documents being dropped

sndth:
This PR fixes documents which hold no node being dropped from the parse result:

```yaml
---
---
foo
```

is parsed as a single document `foo` instead of two documents (`null` and `foo`), and a document made of node properties alone, such as:

```yaml
!!str
```

produces no document at all instead of one holding an empty scalar tagged with `!!str`.

### Cause

`deserialize_directives()` handles `---` by marking the document as begun and keeps reading, so a second `---` right after the first one is consumed as part of the same directives section and the empty document the first one began is never emitted. A `---` found once a document has been begun now ends that document instead.

`deserialize_document()` only counts a document when some token other than the end of the buffer or of the document follows. Node properties are consumed before that check, so a stream holding nothing but properties yields no document, even though the properties belong to an empty scalar. Found properties now count as well, attaching them to the root node was already handled.

### Tests

`Deserializer_MultipleDocuments` gains two subcases: consecutive `---` markers and a document holding nothing but a tag. Two more YAML test suite cases pass: `6XDY`, `UKK6-02` (38 failing cases decreased to 36 with no newly failing case).

---

## Pull Request Checklist

## Please don't

### DIFF
--- a/include/fkYAML/detail/input/deserializer.hpp
+++ b/include/fkYAML/detail/input/deserializer.hpp
@@ -212,8 +212,15 @@ private:
         uint32_t indent = lexer.get_last_token_begin_pos();
         const bool found_props = deserialize_node_properties(lexer, token, line, indent);
 
-        // A stream which only holds comments, white spaces or a bare "..." contains no document.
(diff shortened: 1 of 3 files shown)


fktn-k (approved):

PR #621, opened 2026-09-10
Accept the YAML 1.x minor versions which the library does not know

sndth:
This PR fixes documents being rejected when their `%YAML` directive names a minor version other than 1.1 or 1.2:

```yaml
%YAML 1.3
---
foo
```

throws `parse_error: Only 1.1 and 1.2 can be specified as the YAML version.` instead of being parsed.

### Cause

The lexer accepted exactly `1.1` and `1.2` and rejected everything else. The spec says a processor should reject a document with a different major version, but only warn about a higher minor version and keep parsing it (see [6.8.1. "YAML" Directives](https://yaml.org/spec/1.2.2/#681-yaml-directives)).

The lexer now accepts `1.` followed by one or more digits. A different major version (`2.0`) and malformed values (`1.`, `1.A`, `12`) are still rejected. An unknown minor version is handled as 1.2, which is what `convert_yaml_version()` already falls back to, so `get_yaml_version_type()` returns `VERSION_1_2` for such a document. The library has no way to emit a warning, so none is emitted.

### Tests

`LexicalAnalyzer_YamlVersionDirective` gains a subcase for unknown minor versions (`1.3`, `1.23`, `1.11`), which were moved out of the invalid-value subcase, and the invalid-value subcase gains `1.`. One more YAML test suite case passes: `BEC7` (31 failing cases decreased to 30 with no newly failing case).

---

## Pull Request Checklist

## Please don't

### DIFF
--- a/include/fkYAML/detail/input/lexical_analyzer.hpp
+++ b/include/fkYAML/detail/input/lexical_analyzer.hpp
@@ -800,8 +800,13 @@ private:
 
         m_yaml_version = str_view {m_token_begin_itr, m_cur_itr};
 
-        if FK_YAML_UNLIKELY (m_yaml_version.compare("1.1") != 0 && m_yaml_version.compare("1.2") != 0) {
-            emit_error("Only 1.1 and 1.2 can be specified as the YAML version.");
+        // A YAML processor must reject a different major version, but should only warn about a minor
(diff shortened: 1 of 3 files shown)


fktn-k (approved):

PR #622, opened 2026-09-10
Fix trailing spaces in implicit mapping keys

sndth:
This PR fixes the white spaces in front of a mapping value indicator being kept in an implicit key when the indicator is not followed by a space:

```yaml
foo :
  bar
```

is parsed with the key `"foo "` instead of `"foo"`. The same happens with a tab after the indicator (`foo :<TAB>bar`) and with a flow indicator after it in a flow collection (`{foo :, bar :}`), and `foo :` at the end of the input is parsed as the scalar `"foo :"` instead of a mapping.

### Cause

When scanning a plain scalar, `determine_plain_scalar_range()` ended the scalar in front of ` :` only if the `:` was followed by a space. A `:` followed by a tab, a line break or the end of the input is a mapping value indicator as well, and so is a `:` followed by a flow indicator inside a flow collection (see [7.3.3. Plain Style](https://yaml.org/spec/1.2.2/#733-plain-style)). The scalar now ends in all of these cases, while a `:` followed by a flow indicator outside flow collections (`foo :,bar`) is still part of the scalar.

### Tests

`Deserializer_BlockMapping` gains subcases for a line break, a tab and the end of the input after the indicator, and for
`foo :,bar` staying a scalar. `Deserializer_FlowMapping` gains a subcase for flow indicators after the indicator. One more YAML test suite case passes: `7FWL` (31 failing cases decreased to 30 with no newly failing case).

---

## Pull Request Checklist

## Please don't

### DIFF
--- a/include/fkYAML/detail/input/lexical_analyzer.hpp
+++ b/include/fkYAML/detail/input/lexical_analyzer.hpp
@@ -1240,10 +1240,33 @@ private:
                 case '#':
                     ends_loop = true;
                     break;
-                case ':':
-                    // " :" is permitted in a plain style string token, but not when followed by a space.
-                    ends_loop = (next_pos + 1 < sv.size()) && (sv[next_pos + 1] == ' ');
+                case ':': {
+                    // " :" is permitted in a plain style string token, but not when the ":" is a mapping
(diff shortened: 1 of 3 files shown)


fktn-k (approved):

PR #625, opened 2026-09-12
Keep the tag name when a node is anchored

sndth:
This PR fixes the tag name of a node being lost when an anchor name is added to it afterwards:

```cpp
fkyaml::node node(std::string("foo"));
node.add_tag_name("!!str");
node.add_anchor_name("anchor");

node.has_tag_name(); // false
```

Setting the two in the opposite order keeps the tag, so the result depends on the order in which they are set. A node which is already an anchor and is given another anchor name loses its tag as well.

### Cause

`add_anchor_name()` moves the value of the node into the anchor table and leaves behind a reference to it. The node properties travel with the value, so a tag which was set beforehand ends up in the anchor table while the caller holds the reference, which carries the anchor name alone. A tag set after the anchor name lands on that reference and is therefore kept.

The tag is now taken before the value is moved and set on the reference, which is where it would have been stored had it been set after the anchor name. A node which is anchored again has its tag taken before the previous anchor is resolved, so renaming an anchor keeps the tag too.

Both overloads of `add_anchor_name()` were identical apart from moving the anchor name at the end, and now share one implementation, so the fix applies to both.

### Tests

`Node_AddAnchorNameKeepsTagName` covers a tag set before the anchor name, one set after it, and a node which is given another anchor name.

---

## Pull Request Checklist

## Please don't

### DIFF
--- a/include/fkYAML/node.hpp
+++ b/include/fkYAML/node.hpp
@@ -1221,27 +1221,7 @@ public:
     /// @param[in] anchor_name An anchor name. This should not be empty.
     /// @sa https://fktn-k.github.io/fkYAML/api/basic_node/add_anchor_name/
     void add_anchor_name(const std::string& anchor_name) {
-        if (is_anchor()) {
-            m_attrs &= ~detail::node_attr_mask::anchoring;
-            auto itr = mp_meta->anchor_table.equal_range(anchor_prop()).first;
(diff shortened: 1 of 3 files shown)


fktn-k (approved):
Good catch! The changes look good to me.  
Thanks a lot!

PR #633, opened 2026-09-13
Fix block scalar headers on a line of their own

sndth:
This PR fixes block scalars whose header sits on a line of its own, which currently fail to parse:

```yaml
foo:
  |
  bar
```

```
parse_error: The ":" mapping value indicator must be followed after a mapping key. (at line 3, column 0)
```

The lexer took the base indentation of a block scalar from the line its header sits on. When the header is alone on its line, or only follows node properties there, that line's indentation says nothing about the parent node. So `bar` was treated as less indented, the scalar ended up empty and `bar` was read as a new key. The same happens when node properties push the header down, as in `Spec Example 8.21`:

```yaml
folded:
   !foo
  >1
 value
```

When nothing but node properties precede the header on its line, the base indentation is now taken from the nearest line above that holds more than node properties or a comment. Headers sharing a line with a key, `- ` or `? ` keep their current behavior.

This fixes M5C3 from the YAML test suite.

### Tests
New subcases in `Deserializer_BlockLiteralScalar` and `Deserializer_BlockFoldedScalar`, each failing without the fix:
- a header on a line of its own
- a header after `&anchor !!str` on a line of their own, with `|1`
- a header with `>1` after a tag on a line of its own (the M5C3 case)

---

## Pull Request Checklist

## Please don't


fktn-k (approved):

PR #634, opened 2026-09-13
Reject node properties which are not indented enough

sndth:
This PR makes node properties on a line of their own fail to parse when they are not more indented than the block collection they are in. Currently they are silently accepted and bound to the next node:

```yaml
- item1
&node
- item2
# -> ["item1", &node "item2"]
```

```yaml
seq:
&anchor
- a
# -> {seq: &anchor ["a"]}
```

Properties which end their line belong to a node nested in the current block collection, so they must be indented more than that collection. The check sits in `deserialize_node_properties()`, which every path parsing node properties goes through. Properties which share a line with a key, `- ` or `---` are always more indented than their collection and are not affected, and neither are valid inputs like:

```yaml
seq:
 &anchor
- a
```

This fixes G9HC and GT5M from the YAML test suite.

### Tests
New subcases in `Deserializer_NodeProperties`, each accepted without the fix:
- properties on a line of their own in a block sequence
- properties on a line of their own before a zero indented block sequence
- a tag on a line of its own in a block mapping

Plus a valid case with the properties indented one space more than the mapping.

---

## Pull Request Checklist

## Please don't


fktn-k (approved):
The changes look good to me. Although I wasn't sure if [the last added test case](https://github.com/fktn-k/fkYAML/pull/634/changes#diff-a77c9d885e3900452e6ff77837db424992c40798a7406fc1b35f76d51f45a521R4447) is a valid YAML, I confirmed that all the parsers in the YAML playground accepted it. Thank you for looking into such an edge case.

PR #636, opened 2026-09-14
Reject block collection entries on document start lines

fktn-k:
This PR improves the deserializer so it can reject node properties before a block node on the same line, which was incorrectly regarded as a valid YAML by the deserializer.

## Summary

- Reject node properties before a block sequence entry on the same line.
- Reject node properties before an implicit block mapping on the document start line.
- Preserve valid properties for block collections beginning on following lines.
- Correctly reject invalid entries after consecutive document start markers.
- Add regression tests for anchors, tags, and their combination.

## Testing

- Add regression tests for anchors, tags, and their combination. All the unit test cases pass successfully.
- Fix `CXX2` and `SY6V` in the YAML Test Suite. As a result, 4 failing cases have decreased to 2 with no newly failing case.

---

## Pull Request Checklist

## Please don't


include/fkYAML/detail/input/deserializer.hpp
```diff
@@ -221,6 +223,8 @@ class basic_deserializer {
         uint32_t line = lexer.get_lines_processed();
         uint32_t indent = lexer.get_last_token_begin_pos();
         const bool found_props = deserialize_node_properties(lexer, token, line, indent);
+        m_reject_root_mapping_with_props =
+            m_has_explicit_document_start && found_props && line == m_explicit_document_start_line;
```
sndth: Should a block mapping on the document start line be rejected even without node properties? Currently `--- a: b` is still accepted, although it is invalid for the same reason as `--- &anchor a: b`: a block collection has to begin on a line of its own.

Dropping `found_props` here would cover it (and the flag name would need adjusting). No suite case catches it: 9KBC (`--- key1: value1`) already fails on its second line because of the indentation.

sndth (commented):
Great PR! I did a small code review.

include/fkYAML/detail/input/deserializer.hpp
sndth: The same applies to block sequences: `--- &a - a` is rejected now, but `--- - a` is still accepted.

https://github.com/fktn-k/fkYAML/blob/63080f7fd5bc88b5894ebb267cea6f05373900e5/include/fkYAML/detail/input/deserializer.hpp#L243-L246

So both checks could drop the `found_props` condition. No suite case covers the sequence either.

sndth: Correction: for the sequence check, dropping `found_props` alone is not enough. Without properties `line` is the line of the `- ` itself, so every root sequence like `- a` would be rejected. This works instead:

```cpp
if FK_YAML_UNLIKELY (
    (found_props && line == lexer.get_lines_processed()) ||
    (m_has_explicit_document_start && lexer.get_lines_processed() == m_explicit_document_start_line)) {
```

With that, and `found_props` dropped from `m_reject_root_mapping_with_props`, all unit tests pass and the suite result is unchanged (only 5U3A and 6CK3 fail). The error message would need adjusting too, since `--- - a` has no node properties.

One gap remains in both checks: a document begun by a `---` which ends the previous one (`---` followed by `--- a: b`) is not caught, because the start line is only recorded in `deserialize_directives()`.

fktn-k: @sndth  
Thank you for the review. I couldn't take the mentioned points into consideration.  
I added 2 commits which addresses the points. I'd appreciate it if you could review them.
- Block collection entries on a document start line without node properties are now correctly rejected. As you corrected, just dropping `found_props` from the conditional expressions caused a regression. So I separately check `found_props && line == lexer.get_lines_processed()` and `m_has_explicit_document_start && lexer.get_lines_processed() == m_explicit_document_start_line` with dedicated error messages for better explanation of the syntax errors.
- Both `---\n--- a: b` and `---\n--- - a` are now correctly rejected by modifying the data management so that the line number of `---` consumed in the previous document would be carried over to the next document.

sndth: Looks good to me, thank you! 😸

Two small nits, if you like:
- "a block mapping entry cannot be on the document start line." starts with a lowercase letter, unlike the sequence one.
- The new "invalid root mapping" subcase uses `REQUIRE_THROWS` instead of `REQUIRE_THROWS_AS(..., fkyaml::parse_error)`.

fktn-k: I fixed them. Thanks!

PR #637, opened 2026-09-15
Reject block sequences on the line of a key or properties

sndth:
This PR fixes block sequences starting on the line of a mapping key or of node properties in nested nodes (5U3A in the YAML test suite). A block sequence has to begin on a line of its own, so these inputs are now rejected:

```yaml
foo: !!seq - bar
- &anchor - foo
a: b
c: - d
```

Properties before `-` in nested nodes use the same error message as the root check from #636.

Tests: unit test cases for a mapping key and for properties on a mapping value, a sequence entry and the root.

---

## Pull Request Checklist

## Please don't


fktn-k (approved):
We have just one more test case to fix, then fkYAML will fully cover the YAML test suite!

sndth:
> The changes look good to me. Thanks a lot! We have just one more test case to fix, then fkYAML will fully cover the YAML test suite!

Thanks! About the last one, 6CK3: the spec contradicts itself there. Section 5.6 says processors must not expand %-escaped characters, but Example 6.26 resolves `!e!tag%21` to `tag:example.com,2000:app/tag!`, and the test suite follows the example. PyYAML decodes %xx in tags as well.

Would you like `get_resolved_tag_name()` to decode %xx (keeping `get_tag_name()` as written), or would you rather skip 6CK3 in the runner? I'm happy to open a PR for either, unless you're already working on it.

fktn-k:
> > The changes look good to me. Thanks a lot! We have just one more test case to fix, then fkYAML will fully cover the YAML test suite!
> 
> Thanks! About the last one, 6CK3: the spec contradicts itself there. Section 5.6 says processors must not expand %-escaped characters, but Example 6.26 resolves `!e!tag%21` to `tag:example.com,2000:app/tag!`, and the test suite follows the example. PyYAML decodes %xx in tags as well.
> 
> Would you like `get_resolved_tag_name()` to decode %xx (keeping `get_tag_name()` as written), or would you rather skip 6CK3 in the runner? I'm happy to open a PR for either, unless you're already working on it.

@sndth  
You're right. I've been confused about the contradiction.  
After checking how other YAML libraries process %-escaped characters with the YAML parser playground, I would now like to make this library decode %-escaped characters only for resolved tags with tag shorthands in `get_resolved_tag_name()` since many parser follows the output in the YAML test suite; no modification to the node's `mp_prop->tag` so serialization can easily keep the original %-escaped characters. Also, there should be no decoding for verbatim tags which is clearly stated as `the YAML processor must deliver the verbatim tag as-is to the application`. (interestingly, the reference parser decodes %-escaped characters in verbatim tags too, though)  
I'd really appreciate it if you could open a pull request for it since I don't have any prepared patch for it.

fktn-k:
@sndth  
I would like to revert my opinion, uri decoding in `get_resolved_tag_name()` since I found an unresolved issue in the YAML test suite repository. In the issue, the maintainer says the emitted tag should be uri encoded.  
https://github.com/yaml/yaml-test-suite/issues/9

The result is likely to be a bug in the reference parser. If that is true, the contradiction would be resolved. So, let me decide to leave %-escaped characters as-is and skip the 6CK3 case.

sndth:
@fktn-k 
Makes sense, thanks for digging up yaml/yaml-test-suite#9. Section 5.6 is normative while Example 6.26 is only illustrative, so keeping %-escaped characters as-is is the safer reading.

Would you like me to open a PR which skips 6CK3 in the runner with its own reason (linking to yaml/yaml-test-suite#9), or are you going to do it yourself?

fktn-k:
@sndth  
I would like you to open it if you have time.

PR #638, opened 2026-09-15
Fix empty mapping keys after omitted values and node properties

sndth:
This PR fixes empty mapping keys in block mappings. No yaml-test-suite case covers this; bugs were found while working on the suite.

An empty key which follows node properties on its line, or a key whose value is omitted, was either rejected or silently parsed into a wrong structure:

```yaml
key: value
&anchor : empty key   # was a parse error, now {key: value, &anchor null: empty key}

foo:
!!str : bar           # was {foo: bar}, now {foo: null, !!str "": bar}

foo: &map
  &key : bar          # was {foo: &map bar}, now {foo: &map {&key null: bar}}

foo: &anchor
: bar                 # was a parse error, now {foo: &anchor null, null: bar}

foo:
  : bar               # was {foo: bar}, now {foo: {null: bar}}
```

The condition from #608, which handled only `- &anchor : bar`, is generalized to block mappings and mapping values. A key separator which directly follows the one of a key with an omitted value now begins an entry with an empty key, and properties between the two separators are bound to the omitted value.

Tests: unit test cases for empty keys with properties (after a mapping entry, after an omitted value, in a nested mapping) and for empty keys after an omitted value (with and without properties, on the same and a deeper indentation).

---

## Pull Request Checklist

## Please don't


fktn-k (approved):

PR #513, opened 2026-03-04
Support unsigned 64-bit integers (uint64_t) exceeding INT64_MAX (#501)

sndth:
Large positive decimal scalars whose value exceeds INT64_MAX (e.g. xxHash digests, high-half UUID fields) were silently parsed as STRING nodes instead of INTEGER nodes. Calling as_int() on such a node then threw type_error, making the value unreachable without ugly string post-processing.

This PR adds transparent uint64_t storage support inside the existing int64_t field via a new style flag, uint_bit, and exposes two new public APIs: is_uint() and as_uint().

---

## Pull Request Checklist

Read the [CONTRIBUTING.md](https://github.com/fktn-k/fkYAML/blob/develop/.github/CONTRIBUTING.md) file for detailed information.  

- [x] Changes are described in the pull request or in a referenced [issue](https://github.com/fktn-k/fkYAML/issues).
- [x] The test suite compiles and runs without any error.
- [ ] [The code coverage](https://coveralls.io/github/fktn-k/fkYAML) on your branch is 100%.
- [ ] The documentation is updated if you added/changed a feature.

## Please don't

- The C++11 support varies between different **compilers** and versions. Please note the [list of supported compilers](https://github.com/fktn-k/fkYAML/blob/develop/README.md#supported-compilers). Some compilers like GCC 4.7 (and earlier), Clang 3.3 (and earlier), or Microsoft Visual Studio 13.0 and earlier are known not to work due to missing or incomplete C++11 support. Please refrain from proposing changes that work around these compiler's limitations with `#ifdef`s or other means.
- Please refrain from proposing changes that would **break [YAML](https://yaml.org/) specifications**. If you propose a conformant extension of YAML to be supported by the library, please motivate this extension.
- Please do not open pull requests that address **multiple issues**.

PR #639, opened 2026-09-15
Serialize unsigned 64-bit integers

sndth:
This PR fixes serializing integers above `INT64_MAX`, which #513 made parsable. Such a value is stored with the `uint_bit` flag, but the serializer read it as a signed integer, so serializing it threw a `type_error`. Constructing a node from a `uint64_t` above that range did not set the flag either, so the value became `-1`. `std::hash<fkyaml::node>` had the same problem as the serializer:

```cpp
auto parsed = fkyaml::node::deserialize("a: 18446744073709551615");
fkyaml::node::serialize(parsed);  // threw type_error, now "a: 18446744073709551615\n"

fkyaml::node built = std::numeric_limits<uint64_t>::max();
fkyaml::node::serialize(built);   // was "-1", now "18446744073709551615"

std::hash<fkyaml::node>()(parsed["a"]);  // threw type_error, now hashes the value
```

The serializer and the hash now read flagged integers with `as_uint()`. `to_node` flags an unsigned value which exceeds the range of the node's integer type, choosing that path at compile time so that no always-false comparison is instantiated for narrower types.

Tests: unit test cases for serializing parsed and constructed `uint64_t` values, constructing nodes from `uint64_t` inside and above the signed range, and hashing unsigned and negative integer nodes.

---

## Pull Request Checklist

## Please don't


fktn-k (approved):
Good catch! I didn't know such usecases weren't covered. Thanks a lot!

PR #640, opened 2026-09-15
Skip 6CK3 in the YAML test suite runner

sndth:
This PR skips the yaml-test-suite case 6CK3, the last failing one, so the runner passes every case.

6CK3 (Spec Example 6.26) expects the tag `!e!tag%21` to resolve to `tag:example.com,2000:app/tag!`, i.e. with `%21` expanded. Section 5.6 of the YAML 1.2.2 spec says the processor must not expand such escaped characters, and the expected output is questioned in yaml/yaml-test-suite#9, so fkYAML keeps them as-is as discussed in #637.

The case gets its own skip reason instead of reusing the duplicate key one. Only the YAML variant is skipped, since the JSON variant of 6CK3 passes.

Tests: the YAML test suite runner now passes all cases pass! 🔥 6CK3 being skipped like 2JQS and X38W.

---

## Pull Request Checklist

## Please don't


fktn-k (approved):
I'll merge this into develop once all the workflows pass successfully. (The workflows for clang compilers are unstable...)

PR #643, opened 2026-09-16
Fix empty flow sequence entries with node properties

sndth:
This PR fixes flow collection entries which consist of node properties only. No yaml-test-suite case covers this.

Such an entry was not added to its sequence, so its properties were applied to a wrong node, or the input was rejected:

```yaml
[a, &x]            # was &x ["a"], now ["a", &x null]
[&x, b]            # was a parse error, now [&x null, "b"]
[a, !!str
]                  # was the whole document !!str "", now ["a", !!str ""]
&m
[a, &x
]                  # was &x ["a"], now &m ["a", &x null]
{a: &x
, b: c}            # was {a: null, b: &x c}, now {a: &x null, b: c}
```

There were two causes. Properties which end their line were deferred in the flow context as in the block context, although line breaks do not change the node they belong to there. And a flow sequence entry with properties only never created a node, so the entry is now added as an empty node at the value separator or the sequence end, as omitted flow mapping values already are.

Tests: unit test cases for flow sequence entries with properties only (first, last, a tag ending its line, an anchored sequence) and for flow mapping values with properties which end their line.

---

## Pull Request Checklist

## Please don't


fktn-k (approved):

PR #648, opened 2026-09-18
Bind properties before a flow key to its mapping

sndth:
This PR fixes node properties on a line of their own before a flow collection which turns out to be an implicit mapping key. No yaml-test-suite case covers this.

Such properties belong to the mapping which the key begins, as they already do for scalar keys, but they were applied to the flow collection itself:

```yaml
&mapping
[a]: value         # was {&mapping [a]: value}, now &mapping {[a]: value}

foo: !!map
  {a: b}: c        # was {foo: {!!map {a: b}: c}}, now {foo: !!map {{a: b}: c}}

? &m
  [a]: c
: b                # was {{&m [a]: c}: b}, now {&m {[a]: c}: b}
```

Whether a flow collection is a key is only known after its end, so the deferred properties are now held in the context of the collection and applied at its end: to the mapping if a key separator follows, otherwise to the collection itself. A flow collection which is not a key, or which is an explicit key on its own, keeps its properties as before.

Tests: unit test cases for flow sequence and flow mapping keys after properties, a flow collection which is not a key, and a flow collection as or inside an explicit key.

---

## Pull Request Checklist

## Please don't


fktn-k (approved):
Good catch! Thanks for this great PR.

PR #649, opened 2026-09-19
Compare and convert unsigned 64-bit integers

sndth:
This PR fixes comparisons and conversions of integers above `INT64_MAX`, following #639. Such a value is stored with the bit pattern of a negative integer and a `uint` flag, but `==`, `<` and the conversions to `bool` and floating point numbers ignored the flag:

```cpp
auto uint_max = fkyaml::node::deserialize("18446744073709551615");
auto minus_one = fkyaml::node::deserialize("-1");

uint_max == minus_one;           // was true, now false
minus_one < uint_max;            // was false, now true
uint_max.get_value<double>();    // threw type_error, now 1.8446744073709552e19
uint_max.get_value<bool>();      // threw type_error, now true
```

```yaml
18446744073709551615: a
-1: b                            # was "Detected duplication in mapping keys", now a mapping with 2 entries
```

`==` now also requires the same flag. `<` orders a flagged integer above any signed one and compares two flagged integers as `uint64_t`. The conversions to `bool` and floating point numbers read flagged integers with `as_uint()`.

Tests: unit test cases for equality and ordering of unsigned, signed and negative integers, mapping keys which share a bit pattern, and converting unsigned and negative integers to `bool` and `double`.

---

## Pull Request Checklist

## Please don't


fktn-k (approved):
Thanks for this follow up! The changes look good to me.  
I forgot to consider these operators in the last review. Sorry for that.

sndth:
> Thanks for this follow up! The changes look good to me. I forgot to consider these operators in the last review. Sorry for that.

Thank you, no worries! 😸

PR #650, opened 2026-09-20
Defer properties before an explicit key block value

sndth:
This PR fixes a parse error when the value of an explicit mapping key has node properties and begins as a block collection on the next line:

```yaml
? foo
: &anchor
  - bar
# -> parse_error: Node properties cannot precede a block sequence entry on the same line.
```

A block mapping value fails with "Detected invalid indentation." for the same reason.

The properties after the separator belong to the key's value, which begins on a later line, so the node they bind to is not known at the separator. They were kept pending, and the block sequence entry check rejects any pending properties regardless of their line. The value of a plain mapping key defers them until the node which follows is known, and an explicit key's value now does the same:

```yaml
? foo
: &anchor
  - bar
# -> {foo: &anchor [bar]}
```

Properties on the same line as a block sequence entry (`: &anchor - bar`) are still an error.

Tests: explicit keys with an anchor, a tag, and both before block sequence and block mapping values on the following lines; an anchored value resolved through an alias; same-line properties before a block sequence entry still rejected.

---

## Pull Request Checklist

## Please don't


fktn-k (approved):
Good catch! Thanks for this patch.

